PR #1, opened 2024-06-03
Create basic architecture

jean-michelet:
# Proposal for Basic Fastify API Architecture

This PR proposes a basic architecture for a Fastify API. It does not contain any features. I suggest we open an issue and create a roadmap for the next steps.

The base was generated by (and designed to work with) [fastify-cli](https://github.com/fastify/fastify-cli).

Core plugins used:
`helmet`, `cors`, `swagger`, `swagger-ui`, `env`, `sensible`, `under-pressure`, `mysql`

`fastify-cli` generates a lot of comments, but I decided to keep them because they are educational.
I've also added a `server.ts` file just to show how not to depend on `fastify-cli` to launch the app.

Wait for your feedbacks.


melroy89:
Also I would personally love to see typescript! Together with ESM.

hpmouton:
Is there perhaps a debug bar plugin for fastify which can give us metrics on the performance of our server,similar to the debug bar we see in the symfony demo and laravel framework? I can't seem to find one. 
I think Typescript integration would be great.

On the note of the roadmap what is your initial ideas on the features and use cases of this project?

melroy89:
> I'm not sure the way I've organized the plugins is relevant @melroy89.
> I've tried to leverage the autoloader to keep the code clean, but I'd like your feedback.

Autoloader is good to use indeed! And also the best way of doing it IMO. I personally have my files within the `plugins` directory without the additional sub-directories now you created. Which I believe will reduce the complexity.

I personally added the Swagger `fastify.register` in server.ts. You created a separate plugin for only register the swagger plugin, I believe this is not necessary.

melroy89:
> Is there perhaps a debug bar plugin for fastify which can give us metrics on the performance of our server,similar to the debug bar we see in the symfony demo and laravel framework? I can't seem to find one.

Not that I know of. I do know `@fastify/under-pressure` if you want to limit the calls. And in your case your are looking for `fastify-metrics` maybe to collect data and push it to Prometheus. And then use some Grafana instance for example to plot data.

But again, no built-in "debug bar" web UI element. I think fastify-metrics is the closest. And of course just use the default Pino logging??

jean-michelet:
> Not that I know of. I do know @fastify/under-pressure if you want to limit the calls.

Ho, yep, I should add this plugin too I think.

jean-michelet:
So, todo:

- Adding TypeScript. With Type-Providers I guess.
- Under-pressure plugin, I will link this video of Matteo in the comments: https://www.youtube.com/watch?v=VI29mUA8n9w
- Simplifying plugin folder structure
- Put swagger plugin config directly in app.ts

Do you have any other idea right now for the basic architecture?

server.js
```diff
@@ -0,0 +1,47 @@
...
+  // delay is the number of milliseconds for the graceful close to finish
+  const closeListeners = closeWithGrace({ delay: process.env.FASTIFY_CLOSE_GRACE_DELAY || 500 }, async function ({ signal, err, manual }) {
+    if (err) {
+      app.log.error(err)
+    }
+    await app.close()
+  })
+
+  app.addHook('onClose', (instance, done) => {
+    closeListeners.uninstall()
```
melroy89, line 32: This isn't mentioned on the readme here: https://github.com/mcollina/close-with-grace. I think you are correct with this uninstall I believe you. Unless Matteo is saying this is no longer needed. Anyhow, maybe we should also create a PR for `mcollina/close-with-grace` to update the README.

melroy89:
@jean-michelet You are right on track with [Type Providers](https://fastify.dev/docs/latest/Reference/Type-Providers/) for TS. Especially also for schema validation both out and incoming requests/responses. But also declaring your own version/extending the `FastifyInstance` interface (part of `declare module 'fastify'`) for Fastify config options. 

Maybe also add an common example for authentication using `@fastify/auth`.  Like maybe a simple but yet powerful JWT strategy on top of fastify/auth. Reminder fastify/auth is not providing an authentication strategy itself.

I believe another good practice is to configure `setErrorHandler` on the Fastify app. As well as `setNotFoundHandler`. My code:

```js
    this.app.setErrorHandler((err, req, reply) => {
      req.log.error({ err })
      reply.status(err.statusCode || 500)
      reply.send(err)
    })

    this.app.setNotFoundHandler(async (req, reply) => {
      req.log.info({ req }, 'request was not found')
      reply.code(404)
      return { message: 'Not found' }
    })
```

Maybe you also want to add an caching example. 

Last but not least, best practices for logging. `Fastify()` constructor provides a way to give additional options for the logger, but also `ajv`. Maybe `ajv` is out of scope, but very powerful non the less if you want to let's say remove additional body properties which aren't part of your schema you can configure that via `{ ajv: { customOptions: { removeAdditional: 'all' } }`.  And the logger via: `{ logger: <...>}`. Options for the logger is either a boolean. Or  during development I use a more advanced JS object that I pass to the `logger`:

```js
 {
    level: 'info',
    transport: {
      target: 'pino-pretty',
      options: {
        translateTime: 'HH:MM:ss Z',
        ignore: 'pid,hostname',
      },
    },
  }
 ```
  
 Here a full example, what I have (do whatever you want with this info):
 
 ```js
 const environment = process.env.NODE_ENV || 'production'
  const envToLogger: {
      [key: string]: object | boolean
    } = {
      development: {
        level: 'info',
        transport: {
          target: 'pino-pretty',
          options: {
            translateTime: 'HH:MM:ss Z',
            ignore: 'pid,hostname',
          },
        },
      },
      production: true,
      test: false,
    }
    
    this.app = Fastify({
      logger: envToLogger[environment] ?? true,
      ajv: {
        customOptions: {
          coerceTypes: 'array', // change data type of data to match type keyword
          removeAdditional: 'all',// Remove additional body properties
        },
      },
    }).withTypeProvider<JsonSchemaToTsProvider>()
```

Ps. MAYBE... a MySQL or other database integration example? Like `@fastify/mysql`.

jean-michelet:
I shouldn't have replied by mail, sorry ^^

app.js
```diff
@@ -0,0 +1,27 @@
...
+ * If you would like to turn your application into a standalone executable, look at server.js file
+ *
+ */
+
+import path from 'node:path'
+import AutoLoad from '@fastify/autoload'
+import swagger from './swagger.js'
+
+export default async function app (fastify, opts) {
+  fastify.register(swagger)
```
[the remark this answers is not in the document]
jean-michelet: OK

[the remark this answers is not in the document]
jean-michelet: The doc shows an example with await effectively, I had missed it.

server.js
jean-michelet: Maybe this example need an update so:
https://github.com/fastify/fastify-cli?tab=readme-ov-file#migrating-out-of-fastify-cli-start

app.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * If you would like to turn your application into a standalone executable, look at server.js file
+ *
+ */
+
+import path from 'node:path'
+import AutoLoad from '@fastify/autoload'
```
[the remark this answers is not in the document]
jean-michelet: I think it was generated by default by the cli...

server.js
```diff
@@ -0,0 +1,47 @@
...
+const app = Fastify({
+  logger: true
+})
+
+async function init () {
+  // Register your application as a normal plugin.
+  app.register(appService)
+
+  // delay is the number of milliseconds for the graceful close to finish
+  const closeListeners = closeWithGrace({ delay: process.env.FASTIFY_CLOSE_GRACE_DELAY || 500 }, async function ({ signal, err, manual }) {
```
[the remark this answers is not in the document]
jean-michelet: OK, the signal is not in use at the moment either.

jean-michelet:
@melroy89 

Now that I think about it, maybe I should go beyond the basic architecture and offer a complete application, albeit with a modest scope. It could save review time from Matteo and the Core Team.

I have some reading to do (I wanted to make this demo as an opportunity to learn Fastify deeper), but I am pretty confident I can push a lot to review the next week.

Wdyt?

server.js
jean-michelet: @mcollina

jean-michelet: @mcollina

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+        node-version: ${{ matrix.node-version }}
+        
+    - name: Install dependencies
+      run: npm i --ignore-scripts
+      
+    - name: Lint Code
+      run: npm run lint
+      
+    - name: Test
+      run: npm test
```
Fdawgs, line 31:
```suggestion
name: CI

on:
  push:
    branches:
     - main
     - master
     - next
     - 'v*'
    paths-ignore:
      - 'docs/**'
      - '*.md'
  pull_request:
    paths-ignore:
      - 'docs/**'
      - '*.md'

jobs:
  test:
    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v3
    with:
      license-check: true
      lint: true
```

Please use one of the shared workflows, it will make this easier to maintain going forward.

Fdawgs (changes requested):

.github/workflows/ci.yml
jean-michelet: Do you agree with using ci v4?

Everything is green [here](https://github.com/jean-michelet/demo/pull/2), I used `glob` to run the tests on *windows, Node18*, maybe I should do it differently?

app.js
```diff
@@ -21,4 +21,35 @@ export default async function app (fastify, opts) {
...
+      err,
+      request: {
+        method: request.method,
+        url: request.url,
+        query: request.query,
+        params: request.params
+      }
+    }, 'Unhandled error occurred')
+
+    reply.code(err.statusCode ?? 500)
```
melroy89, line 40: This is a nullish coalescing operator, it only returns 500 if `statusCode` is null. In the case it's just an empty string or a boolean, it will return that value. This is maybe want you want, or maybe not. Just saying.. I use `||` which is an or operator, that returns 500 if `statusCode` is "falsely".

plugins/swagger.js
```diff
@@ -0,0 +1,29 @@
...
+import fastifySwaggerUi from '@fastify/swagger-ui'
+import fastifySwagger from '@fastify/swagger'
+
+export default fp(async function (fastify) {
+  /**
+    * A Fastify plugin for serving Swagger (OpenAPI v2) or OpenAPI v3 schemas
+    *
+    * @see https://github.com/fastify/fastify-swagger
+  */
+  await fastify.register(fastifySwagger, {
```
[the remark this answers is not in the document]
melroy89: I believe so yes. I guess swagger plugin is not depended on other types of autoload plugins. So this might needs to be clarified indeed.

app.js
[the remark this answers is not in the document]
melroy89: Yes, so to extend on this. I had issues with under-pressure plugin when the plugin was loaded synchronize together with my database, causing my `healthCheck` (part of under pressure plugin) to fail during start-up and I needed to wait several seconds. To fix this I added `awaiy` in front of all my `register` methods. As well as the async method I used too bootstrap my app. Which eventually fixed all the synchronize issues. Now my application starts-up within a second, and is fully functional even with the under-pressure with a db healthCheck configured. 

I just wanted to mention this, so you understand my full story.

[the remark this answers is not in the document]
jean-michelet: That's interesting, we may have similar problems in this demo, and this would be an opportunity to think about it and possibly comment on it.

app.js
jean-michelet: Okay, I tend to think that people always use a logicalOr expression check out of habit.
I was considering relying on the expected type here:
```ts
FastifyError.statusCode?: number | undefined
```

jean-michelet:
Should we use [`ts-standard`](https://github.com/standard/ts-standard)?
I use it on my own projects, but the rules are pretty strict. For example, `explicit-function-return-type` is applied.

melroy89:
> Should we use [`ts-standard`](https://github.com/standard/ts-standard)? I use it on my own projects, but the rules are pretty strict. For example, `explicit-function-return-type` is applied.

I'm using eslint and the official typescript-eslint plugin: https://typescript-eslint.io/getting-started/. I use eslint together with `@typescript-eslint/recommended"` and `prettier`.

app.js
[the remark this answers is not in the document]
jean-michelet: Are you allowed to share the code or at least an example that illustrates the problem? That might be useful.
Also, I was wondering if you only configure for the production environment or if you also have test cases?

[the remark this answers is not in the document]
melroy89: I'm afraid that code I do not have anymore, after all I fixed it  :D. After you completed this demo PR, I can of course test if I notice any issue with under-pressure, but as long as you do not setup a DB config with a heathcheck, this will not happen.

.github/workflows/ci.yml
jean-michelet: I'm setting up the database, and I think it's going to be difficult to use shared workflows at the moment. At least, I don't feel competent enough in this area to adapt the existing.

I think I can create a really easy-to-maintain CI here.
There's no need for tons of demos so I think it should be okay, wdyt?

server.js
jean-michelet: PR open [here](https://github.com/mcollina/close-with-grace/pull/14), I resolve the conversation for now.

.github/workflows/ci.yml
jean-michelet: Maybe I could offer a new shared workflow on the repo too, it would be easier to have all the workflows at the same place when updating versions.

melroy89:
I have another general remark; move all the code (except maybe the test code? Leave that in `test`) to a sub-directory called `src`. I personally do not like to have my code in the root-folder next to the README.md, instead I move the code into a separate directory like `src`.  
Update the `package.json` file (or other configs) accordingly if needed.

server.js
melroy89: Matteo is saying it's not needed, which is strange. Meaning the documentation of fastify-cli is incorrect?...

.github/workflows/ci.yml
jean-michelet: Ok, if I think Matteo prefer I create the workflow here: https://github.com/fastify/workflows/issues/133#issuecomment-2154584252

package.json
```diff
@@ -0,0 +1,51 @@
...
+  "description": "The official Fastify demo!",
+  "main": "app.js",
+  "type": "module",
+  "directories": {
+    "test": "test"
+  },
+  "scripts": {
+    "build": "rm -rf dist && tsc",
+    "watch": "npm run build -- --watch",
+    "test": "tap test/**/*.ts",
```
jean-michelet, line 13: I think it can be a good idea to run test with `tap`, it handle transpilation and code coverage for us.

test/helper.ts
```diff
@@ -0,0 +1,34 @@
+// This file contains code that we reuse
+// between our tests.
+
+import { build as buildApplication } from "fastify-cli/helper.js";
```
jean-michelet, line 4: I need to add `.js` extension here, I think this is a type issue to fix on the cli package `fastify-cli/helper.js`

src/app.ts
```diff
@@ -0,0 +1,68 @@
...
+ * If you would like to turn your application into a standalone executable, look at server.js file
+ */
+
+import fastifyAutoload from "@fastify/autoload";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+import { FastifyInstance, FastifyPluginOptions } from "fastify";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
mcollina, line 11: No need for those

tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "fastify-tsconfig",
+  "compilerOptions": {
+    "outDir": "dist",
+    "typeRoots": ["@types", "node_modules/@types"]
+  },
+  "include": ["@types", "src/**/*.ts"]
```
jean-michelet, line 7: Now we have editor error messages in `test` because it is not included anymore.

But it works.

Not sure how to fix this properly.

melroy89: You mean just use:

```
  "include": [
    "src/**/*.ts",
    "test/**/*.ts"
  ],
  ```

File: src/plugins/under-pressure.ts [hunk 12]
```diff
@@ -0,0 +1,48 @@
...
+      } catch (err) {
+        parent.log.error(err, "healthCheck has failed");
+        throw new Error("Database connection is not available");
+      } finally {
+        connection?.release();
+      }
+    },
+    healthCheckInterval: 5000,
+  } satisfies UnderPressureOptions;
+};
```
mcollina, line 30: Move this inside the plugin definition, remove the function.

src/routes/home.ts
```diff
@@ -0,0 +1,25 @@
...
+          }),
+        },
+        tags: ["Home"],
+      },
+    },
+    async function () {
+      return { message: "Welcome to the official fastify demo!" };
+    },
+  );
+};
```
mcollina, line 23: Please add a post route that validates the body.

src/server.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import Fastify from "fastify";
+
+// Import library to exit fastify process, gracefully (if possible)
+import closeWithGrace from "close-with-grace";
+
+// Import your application as a normal plugin.
+import serviceApp from "./app.js";
+
+const environment = process.env.NODE_ENV ?? "production";
```
mcollina, line 16: Please don’t use NODE_ENV for anything. This is a massive antipattern.

https://youtu.be/HMM7GJC5E2o?si=d9idweist4369pBW

tsconfig.json
jean-michelet: Ho, yep for @types.
But if I include `"test/**/*.ts` it will compile test in dist locally, increasing build time.
How could I include without compiling?

test/plugins/support.test.ts
```diff
@@ -0,0 +1,26 @@
...
+// test('support works standalone', (t) => {
+//   t.plan(2)
+//   const fastify = Fastify()
+//   fastify.register(Support)
+//
+//   fastify.ready((err) => {
+//     t.error(err)
+//     assert.equal(fastify.someSupport(), 'hugs')
+//   })
+// })
```
mcollina, line 26: Remove commented out code

mcollina (commented):
you should also add a few more tests

app.ts
```diff
@@ -0,0 +1,68 @@
...
+ * If you would like to turn your application into a standalone executable, look at server.js file
+ */
+
+import fastifyAutoload from "@fastify/autoload";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+import { FastifyInstance, FastifyPluginOptions } from "fastify";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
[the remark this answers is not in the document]
jean-michelet: Thanks!

That's what I had done, but I deleted it to support workflow on Node18 when I was asked to use a shared workflow.

src/routes/home.ts
jean-michelet: I added to `/example`

src/server.ts
jean-michelet: Is this ok?
https://github.com/fastify/demo/pull/1/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595R21

```ts
function getLoggerOptions() {
  if (process.env.LOGGING === "pretty") {
    return {
      level: "info",
      transport: {
        target: "pino-pretty",
        options: {
          translateTime: "HH:MM:ss Z",
          ignore: "pid,hostname",
        },
      },
    };
  }

  return process.env.LOGGING === "default";
}
```

jean-michelet:
Current TODO for setting up basic architecture:
- add doc about docker-composition
- create CI workflow
- achieve better coverage (test under-pressure health check)

If anyone has feedbacks, I'll continue to work on this next week.
 
@metcoder95 maybe?

src/server.ts
```diff
@@ -13,24 +13,29 @@ import closeWithGrace from "close-with-grace";
...
-      target: "pino-pretty",
-      options: {
-        translateTime: "HH:MM:ss Z",
-        ignore: "pid,hostname",
+/**
+ * Do not use NODE_ENV to determine what logger (or any env related feature) to use
+ * @see https://www.youtube.com/watch?v=HMM7GJC5E2o
+ */
+function getLoggerOptions() {
+  if (process.env.LOGGING === "pretty") {
```
mcollina, line 21:
```suggestion
  if (process.stdout.isTTY) {
```

jean-michelet:
In order to keep a simple CI, I think we should only run the tests on Linux.
It's a demo, not a dependency so I think it's ok.
linux-only

README.md
metcoder95, line 1: It will be nice here to develop a bit more about the methodology or decisions made that lead to this repository; that backstory sometimes helps companies to decide wether or not to adopt a given technology (I've faced that in the past)

eslint.config.js
metcoder95, line 1: We should sync this with the `fastify` repository

src/plugins/cors.ts
metcoder95, line 1: Why not install the plugins directly?

src/plugins/mysql.ts
metcoder95, line 1: Wondering why are we installing the plugins behind another plugin?

src/routes/README.md
```diff
@@ -0,0 +1,27 @@
+# Routes Folder
+
+Routes define routes within your application. Fastify provides an
+easy path to a microservice architecture, in the future you might want
+to independently deploy some of those.
```
metcoder95, line 5: Highlighting concepts like `modular monolith` can be of help here

metcoder95 (commented):
Thanks for pushing for this! 🙏 
I left a pair of comments, so far I believe we can maybe simplify it a little and focus a bit more on the documentation that highlights the pieces of fastify.

src/server.ts
```diff
@@ -31,22 +33,25 @@ function getLoggerOptions() {
     };
   }
 
-  return process.env.LOGGING === "default";
+  // Don't forget to configure it with
+  // a truthy value in production
+  return !!process.env.LOGGING;
```
mcollina, line 38: I generically prefer to expose a `LOG_LEVEL` env variable instead to use in the pretty transport or here.

src/plugins/cors.ts
jean-michelet: We want to leverage autoloader and set plugins in their own files, at least this is what is generated from the fastify-cli templates to use `sensible` for example.
I am not sure about the final structure currently, we should load external plugins first, then custom plugins, then routes.
If we need to be sure autoloader will register the plugins in a specific order, it easily achievable with filenames:
- 1-env.ts
- other-plugin.ts

What is your vision?

src/plugins/mysql.ts
jean-michelet: Just to be autoloaded and having access to the parent instance.
But now that you mention it, I could push a PR to allow to autoConfig via a function loader with parent instance:
https://github.com/fastify/fastify-autoload/blob/930e738e9576f9f341ae186b71c95c3dded02108/index.js#L284

src/routes/README.md
jean-michelet: Yep, a lot of text is auto-generated by fastify-cli, I think I am gonna improve it after this demo.

jean-michelet:
>  I left a pair of comments, so far I believe we can maybe simplify it a little and focus a bit more on the documentation that highlights the pieces of fastify.

Yes, I haven't really thought about the documentation yet. I had planned to split the creation of this repository into several PRs to give interested developers the opportunity to contribute. But I think we should not create something too simple, it's a great opportunity to show implementations with advanced parts of Fastify and Node.js.

Regarding the development process, which do you prefer @mcollina @metcoder95? I can also work until version 0 is ready, I'm fine working on my own ;)

src/server.ts
jean-michelet: Do you mean something like this?
``ts
return { level: process.env.LOG_LEVEL ?? "info" };
```

File: src/plugins/external/1-env.ts [hunk 20]
```diff
@@ -31,7 +30,7 @@ const schema = {
   }
 };
 
-const options = {
+export const autoConfig = {
```
jean-michelet, line 33: Leverage `autoConfig` allows to avoid to wrap plugins in `fp` when not necessary, I didn't notice this autoloader feature.

src/plugins/cors.ts
jean-michelet: I've made some updates, see: https://github.com/fastify/demo/pull/1#discussion_r1634375351

src/plugins/external/under-pressure.ts
```diff
@@ -16,7 +16,7 @@ export default fp(
     fastify.register(fastifyUnderPressure, {
       maxEventLoopDelay: 1000,
       maxHeapUsedBytes: 100_000_000,
-      maxRssBytes: 100_000_000,
+      maxRssBytes: 200_000_000,
```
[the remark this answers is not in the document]
mcollina: I would raise it to 1GB.

src/server.ts
mcollina:
```
return { level: process.env.LOG_LEVEL || 'silent' };
```

So the application does not log during tests

src/plugins/mysql.ts
jean-michelet: Pushed a PR:
https://github.com/fastify/fastify-autoload/pull/381

src/server.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  await app.ready();
+
+  // Start listening.
+  app.listen({ port: process.env.PORT ?? 3000 }, (err) => {
+    if (err != null) {
+      app.log.error(err);
+      process.exit(1);
+    }
+  });
```
mcollina, line 74: use await and not a callback.

jean-michelet:
> Regarding the development process, which do you prefer @mcollina @metcoder95? I can also work until version 0 is ready, I'm fine working on my own ;)

If no answer, I will start to create a API.

Not sure what I should use for authentication though... I think it's a bad idea to use JWT by default, challenging to implement securely imo. At the same time, many Node.js tutorials demonstrate this...

metcoder95:
We can keep it on `v0` for a while until we feel comfortable with it.

metcoder95:
>Not sure what I should use for authentication though... I think it's a bad idea to use JWT by default, challenging to implement securely imo. At the same time, many Node.js tutorials use it...

Can be a good default, tho it is important to document that this is just a demo and further security recommendations should be added

test/helper.ts
```diff
@@ -0,0 +1,56 @@
...
+  app.login = login;
+
+  app.injectWithLogin = async (username: string, opts: InjectOptions) => {
+    opts.headers = {
+      ...opts.headers,
+      Authorization: `Bearer ${await app.login(username)}`
+    };
+
+    return app.inject(opts);
+  };
```
jean-michelet, line 50: I wonder if we can decorate the app before `.ready` is called, I think there are several use cases where it might be useful, e.g. mocking some methods to force failure.

jean-michelet: > Can be a good default, tho it is important to document that this is just a demo and further security recommendations should be added

Can be a good opportunity to educate about security too, but I understand that it could be out of scope.

jean-michelet:
> Can be a good default, tho it is important to document that this is just a demo and further security recommendations should be added

Can be a good opportunity to educate about security too, but I understand that it could be out of scope.

jean-michelet:
How do you feel about this kind of composition?

```ts
import fp from 'fastify-plugin';
import { RowDataPacket } from 'mysql2/promise';
import { FastifyInstance } from 'fastify';

type CreateRepository = (tableName: string) => {
  find: <T>(select?: string, where?: Record<string, any>) => Promise<T | null>;
};

export default fp(
  async function (fastify: FastifyInstance) {
    const processRecord = (record: Record<string, any>) => {
      const keys = Object.keys(record);
      const values = Object.values(record);
      const clause = keys.map((key) => `${key} = ?`).join(' AND ');

      return [clause, values] as const;
    };

    // This is an example to demonstrate the power of composition in fastify
    // You might want to use an ORM or a query builder for your project
    const createRepository: CreateRepository = (tableName) => {
      const find = async <T>(select = '*', where = {}) => {
        const [clause, values] = processRecord(where);

        const query = `SELECT ${select} FROM ${tableName} WHERE ${clause}`;

        const [rows] = await fastify.mysql.query<RowDataPacket[]>(query, values);
        if (rows.length < 1) {
            return null;
        }

        return rows[0] as T;
      };

      return {
        find
      };
    };

    fastify.decorate('createRepository', createRepository);
  },
  { name: 'repository', dependencies: ['mysql'] }
);

```

```ts
import fp from "fastify-plugin";

export default fp(
  async function (fastify) {    
    fastify.decorate("authRepository", fastify.createRepository('users'));
  },
  { name: "authRepository", dependencies: ['repository'] }
);
```

test/helper.ts
melroy89: Ow, maybe we need to fix this upstream in that package

README.md
```diff
@@ -1 +1,39 @@
...
\ No newline at end of file
+# Fastify Official Demo
+
+![CI](https://github.com/fastify/demo/workflows/CI/badge.svg)
+[![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat)](https://standardjs.com/)
+
+> :warning: **Please note:** This repository is still under active development.
+
+The aim of this repository is to provide a concrete example of a Fastify application using what are considered best practices by the Fastify community.
+
```
melroy89, line 9: Maybe explain which Fastify plugins are part of this demo?

jean-michelet: I don't think we need to maintain a list of plugins used in the doc. 

To be honest, I'm not really sure what the documentation should contain, although I'm aware that I should be working on it. IMO, I think it should be very minimalist, explaining the purpose of the demo and how to configure the application. The purpose of the demo is to show concrete code, so people should look at the code and comments.

jean-michelet:
I'm working on the design of the api, a basic task application. 
I'd like to manage authorization with user permissions.
I want to add the ability to add avatars to users and task generation in CSV format to show how to manipulate files.

This could be a good start, I haven't thought much about what to do next, but do you have any ideas? Any plugin suggestions?

README.md
melroy89: OK then not

scripts/seed-database.js
```diff
@@ -0,0 +1,60 @@
...
+      VALUES (?, ?)
+    `
+
+    await connection.execute(insertUserQuery, [username, hash])
+  }
+
+  console.log('Users have been seeded successfully.')
+}
+
+seed()
```
jean-michelet, line 60: It sounds silly, but I can't figure out how to ignore this file from code-coverage.

test/helper.ts
jean-michelet: Yep, there are several other things I want to update on this package. See https://github.com/fastify/demo/pull/1#discussion_r1644008140

I think I am gonna work on it before continuing the demo.

mcollina (approved):
lgtm

I think this is almost in a good solid state!

jean-michelet:
For now, the tests will not pass.
Can you activate the CI @mcollina?

I identified some stuff to fix, but a merge doesn't involve official communication, and it would allow others to contribute.

jean-michelet:
@fastify/collaborators 

A code review would be appreciated.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: npm run lint
+
+      - name: Generate JWT Secret
+        id: gen-jwt
+        run: |
+          JWT_SECRET=$(openssl rand -hex 32)
+          echo "JWT_SECRET=$JWT_SECRET" >> $GITHUB_ENV
+
+      - name: Generate dummy .env for scripts using -env-file=.env flag
+        run: touch .env
```
jean-michelet, line 60: I can also use dotenv to avoid this step, or do you prefer something else?

README.md
jean-michelet: OK, but it would be nice if, even if I'm the one who suggested it, other developers could contribute their perspectives.

src/app.ts
```diff
@@ -0,0 +1,78 @@
...
+    options: { ...opts }
+  });
+
+  // This loads all your custom plugins defined in plugins/custom
+  // those should be support plugins that are reused
+  // through your application
+  fastify.register(fastifyAutoload, {
+    dir: path.join(import.meta.dirname, "plugins/custom"),
+    options: { ...opts }
+  });
```
jean-michelet, line 26: I've suggested this way of doing things, but it remains my intuition.

Also, I wonder if I should add a `priority` export to autoloader, to sort the tree by plugin priority?

```ts

export const autoConfig = { }
export const pluginPriority = 1
export default env;
```

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install dependencies
-        run: npm i --ignore-scripts
```
jean-michelet: Bcrypt related.

src/app.ts
jean-michelet: Although this is easily achievable with file naming.

jean-michelet:
[From metcoder95, regarding the doc](https://github.com/fastify/demo/pull/1/files#r1632821867):
>  It will be nice here to develop a bit more about the methodology or decisions made that lead to this repository; that backstory sometimes helps companies to decide wether or not to adopt a given technology (I've faced that in the past)

OK, but it would be nice if, even if I'm the one who suggested it, other developers could share their perspectives.

scripts/seed-database.js
```diff
@@ -0,0 +1,61 @@
...
+      await connection.query('SET FOREIGN_KEY_CHECKS = 1')
+    }
+  }
+}
+
+async function seedUsers (connection) {
+  const usernames = ['basic', 'moderator', 'admin']
+
+  for (const username of usernames) {
+    const hash = await bcrypt.hash('password', 10)
```
climba03003, line 48: Use `scrypt` instead, so you don't need external dependency.
See https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#introduction

src/schemas/auth.ts
```diff
@@ -0,0 +1,8 @@
+import { Static, Type } from "@sinclair/typebox";
+
+export const CredentialsSchema = Type.Object({
+  username: Type.String(),
+  password: Type.String()
+});
+
+export interface IAuth extends Static<typeof CredentialsSchema> {}
```
climba03003, line 8: Most of the time, I will override the types when using TypeBox.
So, the schema will match what the expected types.

```suggestion

export type CredentialsSchema = Static<typeof CredentialsSchema>
export const CredentialsSchema = Type.Object({
  username: Type.String(),
  password: Type.String()
});
```

test/helper.ts
climba03003: It is the correct way for ESM, `fastify-cli` do not provide exports mapping. So, extension-less will not work.

scripts/seed-database.js
jean-michelet: Can you provide (or give a link to) what you consider a safe implementation please?

src/schemas/auth.ts
jean-michelet: The problem is that `@typescript-eslint/no-redeclare` is currently enforced

src/plugins/custom/repository.ts
```diff
@@ -0,0 +1,103 @@
+import { MySQLPromisePool } from "@fastify/mysql";
+import { FastifyInstance } from "fastify";
+import fp from "fastify-plugin";
+import { RowDataPacket, ResultSetHeader } from "mysql2";
+
+export type IRepository = MySQLPromisePool & ReturnType<typeof createRepository>;
```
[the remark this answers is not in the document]
jean-michelet: I don't have strong opinions on this, if others are okay, I am too.

@types/fastify/index.d.ts
```diff
@@ -0,0 +1,25 @@
+import { MySQLPromisePool } from "@fastify/mysql";
```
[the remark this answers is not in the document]
jean-michelet: This is smart, but I have an augmentation issue if I try to to set user type in jwt plugin:
```
Subsequent property declarations must have the same type.  Property 'user' must be of type 'string | object | Buffer', but here has type 'IAuth'.ts(2717)
jwt.d.ts(37, 5): 'user' was also declared here.
```

So I kept the global file for now, maybe you have a solution for this.

scripts/seed-database.js
climba03003: No link provided, but it is what I currently using with some modify.
For what I use, I store also the scrypt parameter, so I can migrate the old user to safer constant when time pass.
Here is using constant for simple usage.

I choose the second config, with 2 parallelism.
https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt

```js
import { scrypt, timingSafeEqual, randomBytes } from 'crypto'

// Scrypt constants
const SCRYPT_KEYLEN = 32
const SCRYPT_COST = 65536
const SCRYPT_BLOCK_SIZE = 8
const SCRYPT_PARALLELIZATION = 2
const SCRYPT_MAXMEM = 128 * SCRYPT_COST * SCRYPT_BLOCK_SIZE * 2

function hash(value) {
  // Salt must be at least 16 bytes long
  const salt = randomBytes(Math.min(16, SCRYPT_KEYLEN / 2))
  return new Promise((resolve, reject) => {
    scrypt(value, salt, SCRYPT_KEYLEN, { 
      cost: SCRYPT_COST,
      blockSize: SCRYPT_BLOCK_SIZE,
      parallelization: SCRYPT_PARALLELIZATION,
      maxmem: SCRYPT_MAXMEM
    }, function (error, key) {
      if (error !== null) reject(error)
      resolve(`${salt.toString('hex')}.${key.toString('hex')}`)
    })
  })
}

function compare(value, hash) {
  const [salt, hashed] = hash.split('.')
  const buf = Buffer.from(hashed, 'hex')
  new Promise((resolve) => {
    scrypt(value, salt, SCRYPT_KEYLEN, { 
      cost: SCRYPT_COST,
      blockSize: SCRYPT_BLOCK_SIZE,
      parallelization: SCRYPT_PARALLELIZATION,
      maxmem: SCRYPT_MAXMEM
    }, function (error, key) {
      if (error !== null) {
        timingSafeEqual(buf, buf)
        resolve(false)
      } else {
        resolve(timingSafeEqual(key, buf))
      }
    })
  })
}
```

jean-michelet: Interesting, 

I think it would be worthwhile to create an opensource plugin for this.

jean-michelet: Can you take a look? I am not big on encryption stuff: https://github.com/fastify/demo/pull/1/commits/776086b5c5c7c877067a8488abafb2f2210b18c0

src/plugins/custom/scrypt.ts
```diff
@@ -0,0 +1,69 @@
...
+  return new Promise((resolve) => {
+    scrypt(value, saltBuffer, SCRYPT_KEYLEN, { 
+      cost: SCRYPT_COST,
+      blockSize: SCRYPT_BLOCK_SIZE,
+      parallelization: SCRYPT_PARALLELIZATION,
+      maxmem: SCRYPT_MAXMEM
+    }, function (error, key) {
+      /* c8 ignore start - Requires extreme or impractical configuration values */
+      if (error !== null) {
+        resolve(false)
```
climba03003, line 54: Please do not remove the timingSafeEqual here, it keep all the branches use the same time for computation.

src/schemas/auth.ts
jean-michelet: I prefer to have confirmation from other contributors before proceeding, but I have nothing against it.

climba03003 (approved):

scripts/seed-database.js
metcoder95: Maybe just set a wildcard to ignore everything under `scripts/*`?

src/plugins/custom/repository.ts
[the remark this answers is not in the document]
metcoder95: +1 on @climba03003's comment, specially for simplicity; if when adopting others wants to follow that path its fine, but it will be better if we keep it simple for the demo

metcoder95 (approved):
lgtm, thanks for the contribution!
Just left small comments

jean-michelet:
I fixed the CI: https://github.com/jean-michelet/demo/actions/runs/9722290349/job/26835992622

Waiting for the merge or further review before continuing on this.
ci-base-architecture

jean-michelet:
> Maybe just set a wildcard to ignore everything under scripts/*

Don't understand how to it.
I'm having rather a hard time with the Tap documentation, if someone would be kind enough to copy/paste the solution for me, I've spent too much time on it already.

src/routes/README.md
jean-michelet: https://github.com/fastify/fastify-cli/pull/736

jean-michelet:
I worked locally to add a rate limiter, but I think we should merge this before, otherwise I'll make your reviews obsolete.

@mcollina

mcollina (approved):
lgtm

melroy89:
@Fdawgs you are now blocking the merge (request for change) ;). Is there still something we need to fix?

melroy89:
Ah no problem, thanks for your heads up.

src/plugins/external/cors.ts
```diff
@@ -0,0 +1,12 @@
+import cors from "@fastify/cors";
+
+export const autoConfig = {
+  // Set plugin options here
+};
+
+/**
+ * This plugins enables the use of CORS.
+ *
+ * @see https://github.com/fastify/fastify-cors
```
Fdawgs, line 10:
```suggestion
 * @see {@link https://github.com/fastify/fastify-cors}
```

src/plugins/external/helmet.ts
```diff
@@ -0,0 +1,12 @@
+import helmet from "@fastify/helmet";
+
+export const autoConfig = {
+  // Set plugin options here
+};
+
+/**
+ * This plugins sets the basic security headers.
+ *
+ * @see https://github.com/fastify/fastify-helmet
```
Fdawgs, line 10:
```suggestion
 * @see {@link https://github.com/fastify/fastify-helmet}
```

src/plugins/external/1-env.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  // Source for the configuration data
+  // Optional, default: process.env
+  data: process.env
+};
+
+/**
+ * This plugins helps to check environment variables.
+ *
+ * @see https://github.com/fastify/fastify-env
```
Fdawgs, line 78:
```suggestion
 * @see {@link https://github.com/fastify/fastify-env}
```

src/plugins/external/sensible.ts
```diff
@@ -0,0 +1,12 @@
+import sensible from "@fastify/sensible";
+
+export const autoConfig = {
+  // Set plugin options here
+};
+
+/**
+ * This plugin adds some utilities to handle http errors
+ *
+ * @see https://github.com/fastify/fastify-sensible
```
Fdawgs, line 10:
```suggestion
 * @see {@link https://github.com/fastify/fastify-sensible}
```

src/plugins/external/swagger.ts
```diff
@@ -0,0 +1,30 @@
+import fp from "fastify-plugin";
+import fastifySwaggerUi from "@fastify/swagger-ui";
+import fastifySwagger from "@fastify/swagger";
+
+export default fp(async function (fastify) {
+  /**
+   * A Fastify plugin for serving Swagger (OpenAPI v2) or OpenAPI v3 schemas
+   *
+   * @see https://github.com/fastify/fastify-swagger
```
Fdawgs, line 9:
```suggestion
   * @see {@link https://github.com/fastify/fastify-swagger}
```

src/plugins/external/swagger.ts
```diff
@@ -0,0 +1,30 @@
...
+        description: "The official Fastify demo API",
+        version: "0.0.0"
+      }
+    }
+  });
+
+  /**
+   * A Fastify plugin for serving Swagger UI.
+   *
+   * @see https://github.com/fastify/fastify-swagger-ui
```
Fdawgs, line 25:
```suggestion
   * @see {@link https://github.com/fastify/fastify-swagger-ui}
```

src/plugins/external/under-pressure.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+    healthCheckInterval: 5000
+  };
+};
+
+/**
+ * A Fastify plugin for mesuring process load and automatically
+ * handle of "Service Unavailable"
+ *
+ * @see https://github.com/fastify/under-pressure
```
Fdawgs, line 36:
```suggestion
 * @see {@link https://github.com/fastify/under-pressure}
```

src/plugins/external/under-pressure.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Video on the topic: Do not thrash the event loop
+ * @see https://www.youtube.com/watch?v=VI29mUA8n9w
```
Fdawgs, line 39:
```suggestion
 * @see {@link https://www.youtube.com/watch?v=VI29mUA8n9w}
```

src/server.ts
```diff
@@ -0,0 +1,77 @@
...
+
+// Import library to exit fastify process, gracefully (if possible)
+import closeWithGrace from "close-with-grace";
+
+// Import your application as a normal plugin.
+import serviceApp from "./app.js";
+
+/**
+ * Do not use NODE_ENV to determine what logger (or any env related feature) to use
+ * @see https://www.youtube.com/watch?v=HMM7GJC5E2o
```
Fdawgs, line 19:
```suggestion
 * @see {@link https://www.youtube.com/watch?v=HMM7GJC5E2o}
```

src/plugins/external/cors.ts
Fdawgs, line 1: Why aren't we just directly importing these plugins into `src/app.js` like `@fastify/autoload`?

Fdawgs, line 1: I think you need to explicitly register this inside each route and define the `methods` option to ensure each route returns the correct values in the `Access-Control-Allow-Methods` header.

Fdawgs (changes requested):
d

src/plugins/external/cors.ts
jean-michelet: > I think you need to explicitly register this inside each route
Not sure to understand what you mean, the plugin is designed to be configured globally.

src/plugins/external/cors.ts
```diff
@@ -1,12 +1,12 @@
-import cors from "@fastify/cors";
+import cors, { FastifyCorsOptions } from "@fastify/cors";
 
-export const autoConfig = {
-  // Set plugin options here
+export const autoConfig: FastifyCorsOptions = {
+  methods: ['GET', 'POST', 'PUT', 'DELETE']
 };
```
jean-michelet: I'll do a more precise configuration when we will set up the front end, cf #2

Fdawgs (approved):

melroy89:
Nice. It's approved! 🙏🎉. Well done @jean-michelet

jean-michelet:
Can it be merged plz?
I would like to continue.

Issue #15, opened 2024-08-17
CRUD tasks operations


jean-michelet:

jean-michelet:
Use the workflow component as en exprimentation, read: https://github.com/fastify/demo/issues/2#issuecomment-2291693844

PR #16, opened 2024-08-17
Task crud

jean-michelet:
Implement #15


src/app.ts
```diff
@@ -10,6 +10,8 @@ export default async function serviceApp(
   fastify: FastifyInstance,
   opts: FastifyPluginOptions
 ) {
+  delete opts.skipOverride // This option only serves testing purpose
```
jean-michelet, line 13: As I feared, I think we've introduced a problem here: https://github.com/fastify/fastify-cli/pull/742/files#r1689541348

For example, `@fastify/mysql` [passes all the given options directly to `mysql2`](https://github.com/fastify/fastify-mysql/blob/10a2a8e34911d1982ad3556b03424c276e7ccdcf/index.js#L62), which triggers warnings from `mysql2`:
```
Ignoring invalid configuration option passed to Connection: skipOverride. This is currently a warning, but in future versions of MySQL2, an error
will be thrown if you pass an invalid configuration option to a Connection
```

Maybe it's also a bad design to propagate all the options in the autoloader? Like the cli generated template suggest to do:
```
await fastify.register(fastifyAutoload, {
    dir: path.join(import.meta.dirname, "plugins"),
    options: { ...opts }
  });
```

jean-michelet:
If you want to take a look @melroy89

melroy89:
ow you closed it? I'm still on holiday, sorry I didn't look right away

jean-michelet:
I made a mistake, it has been closed automatically, I'll reopen a new one.

PR #19, opened 2024-08-25
feat: crud task according to workflow

jean-michelet:
Issue #15 

I created a task CRUD with workflow configuration.


package.json
```diff
@@ -8,11 +8,12 @@
     "test": "test"
   },
   "scripts": {
-    "build": "rm -rf dist && tsc",
-    "watch": "npm run build -- --watch",
+    "start": "npm run build && fastify start -l info dist/app.js",
+    "build": "tsc",
+    "watch": "tsc -w",
+    "dev": "npm run build && concurrently -k -p \"[{name}]\" -n \"TypeScript,App\" -c \"yellow.bold,cyan.bold\" \"npm:watch\" \"npm:dev:start\"",
+    "dev:start": "fastify start --ignore-watch=.ts$ -w -l info -P dist/app.js",
```
jean-michelet, line 15: Improve the dev mode experience. Inspired by: https://github.com/fastify/fastify-cli/blob/master/generate.js#L54

src/app.ts
```diff
@@ -10,6 +10,7 @@ export default async function serviceApp(
   fastify: FastifyInstance,
   opts: FastifyPluginOptions
 ) {
+  delete opts.skipOverride // This option only serves testing purpose
```
jean-michelet, line 13: As I feared, I think we've introduced a problem here: https://github.com/fastify/fastify-cli/pull/742/files#r1689541348

For example, `@fastify/mysql` [passes all the given options directly to `mysql2`](https://github.com/fastify/fastify-mysql/blob/10a2a8e34911d1982ad3556b03424c276e7ccdcf/index.js#L62), which triggers warnings from `mysql2`:
```
Ignoring invalid configuration option passed to Connection: skipOverride. This is currently a warning, but in future versions of MySQL2, an error
will be thrown if you pass an invalid configuration option to a Connection
```

Maybe it's also a bad design to propagate all the options in the autoloader? Like the cli generated template suggest to do:
```ts
export default async function serviceApp(
  fastify: FastifyInstance,
  opts: FastifyPluginOptions
) {
  await fastify.register(fastifyAutoload, {
      dir: path.join(import.meta.dirname, "plugins"),
      options: { ...opts } // opts propagated here
  });
  
  // code...
}
```

src/plugins/custom/workflow.ts
```diff
@@ -0,0 +1,33 @@
+import fp from 'fastify-plugin';
+import { MultiOriginTransition, Transition, Workflow } from '@jean-michelet/workflow';
```
jean-michelet, line 2: This is a component I created inspired by [symfony/workflow](https://symfony.com/doc/current/components/workflow.html). You can read the documentation here, I am willing to create an official Fastify plugin if you are interested.

src/schemas/tasks.ts
```diff
@@ -1,6 +1,55 @@
...
+} as const;
+
+export const TaskStatus = {
+  New: 'new',
+  InProgress: 'in-progress',
+  OnHold: 'on-hold',
+  Completed: 'completed',
+  Canceled: 'canceled',
+  Archived: 'archived'
+} as const;
```
jean-michelet, line 19: A technic to create safer enums.

jean-michelet:
@melroy89 :wave:

src/app.ts
melroy89: Not sure what to say here, but you are right. Passing all default options that results into this error, is NOT good.

melroy89:
Sorry about my ignorance, but why do we need all those workflow code in a demo app? The CRUD code all seems fine to me. But what does all that workflow code do? Which has even an dependency outside this demo app with `@jean-michelet/workflow`.

Why not just keep the demo clean, and focus on the CRUD part only? Or do I miss something.

jean-michelet:
@gurgunday @climba03003 
Any feedback?

src/plugins/custom/workflow.ts
melroy89: I still thing, this shouldn't be part of the Demo repo.

climba03003 (commented):
IMO, the demo should follow some existing well defined schema.
For example https://petstore.swagger.io/
It shows `authorization`, `linkage between entity`, `multipart handling`, etc.

jean-michelet:
I can take inspiration from it, but maybe we can stick to a task application. I can complete this PR with users endpoints, I just wanted a feedback on a pure CRUD. Autorisation is for a next step.

jean-michelet:
TODO:
- Add users endpoint

mcollina (approved):
lgtm

mcollina (approved):
lgtm

jean-michelet:
Can do Users on an other PR.

README.md
```diff
@@ -41,11 +41,6 @@ To build the project:
 npm run build
 ```
 
-In dev mode, you can use:
-```bash
-npm run watch
```
melroy89, line 46: is watch not working or?

jean-michelet: It does work, but will people use it now that we run it concurrently with the dev server?

src/app.ts
```diff
@@ -6,10 +6,21 @@ import path from "node:path";
 import fastifyAutoload from "@fastify/autoload";
 import { FastifyInstance, FastifyPluginOptions } from "fastify";
 
+export const options = {
+  ajv: {
+    customOptions: {
+      coerceTypes: "array", 
+      removeAdditional: "all",
+      allErrors: true
```
climba03003, line 14: I don't think we should enable `allErrors`.

climba03003 (approved):
I still think that we should follow some existing example.
But since you are the companion of this repository, your motivation on continuing the work is more important.

The only concern is about the `ajv.allErrors`, others code LGTM.

gurgunday (approved):
lgtm

PR #31, opened 2024-09-21
feat: handle user permissions

jean-michelet:
Implements #29 

I don't know if there's a popular way to manage role-based authorizations with Fastify. 
So I made a simple custom implementation.

PR #33, opened 2024-10-09
V0

jean-michelet:
#31 #10 #28 #34

The aim of this PR is to produce enough features to release a v0 (backend only). 
I think it's simpler launch it this way, as the code changes a lot and it's easier to get feedback loop.


imports/ts-node.js
```diff
@@ -0,0 +1,3 @@
+import { register } from 'node:module'
+import { pathToFileURL } from 'node:url'
+register('ts-node/esm', pathToFileURL('./'))
```
jean-michelet, line 3: @mcollina 

I have a warning from node when I try to use experimental loader.
The warning message advise to use this code instead, how would you use ts-node as part of a node command?
Example:
```
node --import ./imports/ts-node.js --env-file=.env ./scripts/seed-database.ts
```

src/routes/api/auth/index.ts
```diff
@@ -24,22 +21,30 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+          const roles = await fastify.knex<{ name: string }>('roles')
+            .select('roles.name')
+            .join('user_roles', 'roles.id', '=', 'user_roles.role_id')
+            .join('users', 'user_roles.user_id', '=', 'users.id')
+            .where('users.username', username)
+
+          const token = fastify.jwt.sign({
+            username: user.username,
+            roles: roles.map((role) => role.name)
+          })
```
jean-michelet, line 41: It does look more performant to put the roles in the jwt than doing an sql query at each request.
But it might have a drawback, e.g. I want a user not being an admin anymore, It could take some time to apply the changes (JWT expiration time).

mcollina (commented):
good work! What's missing?

jean-michelet:
> good work! What's missing?

### To do for a v0
- Use tsx
- I think we should add a pagination to the tasks
- I think we should be able to upload task thumbnails

People that might be give a feedback: @gurgunday @climba03003 @melroy89 @metcoder95 @melroy89

### Maybe for the future?
Initially, I wanted more features like:
- User API for Profile and Password Management
- Registration form with email confirmation

But, although useful from a pedagogical perspective, take the focus away from Fastify and its core plugins. Another problem is that more features leads to more work and maintenance, and I understand now that I will not have the help I expected to have in terms of contribution on this repo. So I think it's better to have a small, quality demo repo that I can commit to maintaining over the long term, than a more ambitious project that doesn't age well.

### Frontend
I also wanted to build a frontend with Fastify/Vite, there is [this PR](https://github.com/fastify/demo/pull/21) for it,  but I am not really confident with the TypeScript support of Fastify/Vite for now.

### Fastify typing
I would like to work on improving TypeScript on Fastify to improve this demo, to get rid of merging declaration, but I see a lot of different PRs and opinions on the subject. 

### CLI DX
There is also [this PR](https://github.com/fastify/fastify-cli/pull/743) from @climba03003 we should review to offer a better CLI experience.

melroy89:
> People that might be give a feedback: @gurgunday @climba03003 @melroy89 @metcoder95 @melroy89

Twice me name 😆 . I'm just one person btw.

imports/ts-node.js
melroy89: Yea it used to be a big issue actually.. I needed to apply various workarounds in order to get loader working in Fastify:

```sh
FASTIFY_AUTOLOAD_TYPESCRIPT=1 VITEST=true node --no-warnings=ExperimentalWarning --loader ts-node/esm src/index.ts
```

I would like to see official support for `ts-node/esm`.

melroy89: Update: Ow it seems that `FASTIFY_AUTOLOAD_TYPESCRIPT=1 VITEST=true ` is no longer required anymore to get it working!? That is cool!

jean-michelet:
Update: following a post on my LinkedIn, I have a lot of people reaching me to contribute on this repo. So there is a chance that we can add more features after this PR and getting more DX feedbacks from the community 🤞

imports/ts-node.js
climba03003: >  I'm now sponsoring you via GitHub!

I don't even notice it is available now. Applied long time ago and forgotten about it.
Thanks for your sponsor.

> Why wouldn't you recommend ts-node/esm?

I am also using `ts-node/esm` in my environment, but it seems like the loader do not get much attention from the `TypeScript` team.

src/plugins/external/session.ts
```diff
@@ -0,0 +1,32 @@
...
+export const autoConfig = (
+  fastify: FastifyInstance
+): FastifySessionOptions => ({
+  secret: fastify.config.COOKIE_SECRET,
+  cookieName: fastify.config.COOKIE_NAME,
+  cookie: {
+    secure: fastify.config.COOKIE_SECURED,
+    httpOnly: true,
+    maxAge: 1800000 // 30 minutes
+  }
```
metcoder95, line 21: I'd advice to also make this one configurable

src/routes/api/autohooks.ts
```diff
@@ -1,9 +1,13 @@
 import { FastifyInstance } from 'fastify'
 
 export default async function (fastify: FastifyInstance) {
-  fastify.addHook('onRequest', async (request) => {
-    if (!request.url.startsWith('/api/auth/login')) {
-      await request.jwtVerify()
+  fastify.addHook('onRequest', async (request, reply) => {
```
metcoder95, line 4: I know this is from autoload experience, but it wouldn't be better to put this as a plugin that can be added on-demand; it seems quite counterintuitive the way is set.

Having the possibility of add this to the routes an implementer wants to protect seems to have better ergonomics than a soft check. wdyt?

metcoder95 (commented):
Overall , LGTM, just left few comments

src/routes/api/autohooks.ts
jean-michelet: Do you recommend to perform this check to every route that need authentication? A set of controllers?

If yes, everything under `/api` requires to be authenticated anyway, so I think it's much safer to have this check here and protect a large part of the application.

src/plugins/external/session.ts
jean-michelet: Ok

src/routes/api/autohooks.ts
jean-michelet: I can add a check here: [src/plugins/custom/authorization.ts](https://github.com/fastify/demo/pull/33/files#diff-00db5aed3aea31be2f7e173651cc17e6984158fc638ae2be571df33a65f56d3b)

metcoder95: Don't have an opinion on that, I believe that it varies from project to project, org to org, developer to developer; nevertheless having an example of that that showcases how to protect on a per-route basis will be of help for the community as they might face this requirement and require for them to apply this logic.

jean-michelet: I have already a per-route approach for authorization, but for authenticated users with roles, isn't it enough?

src/plugins/custom/authorization.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * The use of fastify-plugin is required to be able
+ * to export the decorators to the outer scope
+ *
+ * @see {@link https://github.com/fastify/fastify-plugin}
+ */
+export default fp(
+  async function (fastify) {
+    fastify.decorate('isModerator', isModerator)
+    fastify.decorate('isAdmin', isAdmin)
```
mcollina, line 34: why arent' we doing `request.isAdmin()` instead?

src/plugins/external/cookie.ts
```diff
@@ -0,0 +1,11 @@
...
+import fp from 'fastify-plugin'
+
+/**
+ * This plugins enables the use of cookies.
+ *
+ * @see {@link https://github.com/fastify/fastify-cookie}
+ */
+export default fp(fastifyCookie, {
+  name: 'cookies'
+})
```
mcollina, line 11: Is this file needed at all? Can we do it together with `session`?

src/routes/api/auth/index.ts
```diff
@@ -24,22 +22,29 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+            .join('user_roles', 'roles.id', '=', 'user_roles.role_id')
+            .join('users', 'user_roles.user_id', '=', 'users.id')
+            .where('users.username', username)
 
-          return { token }
+          request.session.user = { username, roles: roles.map(role => role.name) }
+
+          await request.session.save()
+
+          return { success: true }
```
mcollina, line 43: I think this should use a transaction

src/routes/api/tasks/index.ts
```diff
@@ -1,31 +1,50 @@
...
+      }
+
+      const tasksQuery = baseQuery.clone()
+        .select('*')
+        .limit(q.limit).offset(offset).orderBy('created_at', q.order)
+
+      const [tasks, [{ count }]] = await Promise.all([
+        tasksQuery,
+        baseQuery.clone().count({ count: '*' })
+      ])
```
mcollina, line 42: This is a bad practice and it's inefficient. Maybe use some custom SQL?

jean-michelet: ok

src/plugins/external/cookie.ts
jean-michelet: We can

src/plugins/custom/authorization.ts
jean-michelet: Hum, now that I am thinking about it.
I am not sure.

src/routes/api/tasks/index.ts
climba03003: Maybe, 
```sql
SELECT *, COUNT(*) OVER() AS rowNum
FROM tasks
WHER ...
```
Not sure how it goes into knex.

jean-michelet: The problem is that it put the row (`total`) in every task:
```ts
 {
    id: 95,
    name: 'Task 3',
    author_id: 21,
    assigned_user_id: null,
    status: 'completed',
    created_at: 2024-10-18T06:02:07.000Z,
    updated_at: 2024-10-18T06:02:07.000Z,
    total: 89
  },
  {
    id: 94,
    name: 'Task 2',
    author_id: 20,
    assigned_user_id: 21,
    status: 'in-progress',
    created_at: 2024-10-18T06:02:07.000Z,
    updated_at: 2024-10-18T06:02:07.000Z,
    total: 89
  }
// ...
```

jean-michelet: Can you take a look at [this commit](https://github.com/fastify/demo/pull/33/commits/da0fb0da7a710e079d0a86a2ae3eeeb5d90f919e) plz?

src/routes/api/auth/index.ts
jean-michelet:
```ts
async function (request, reply) {
      const { username, password } = request.body

      return fastify.knex.transaction(async (trx) => {
        try {
          const user = await fastify
            .knex<Credentials>('users')
            .select('username', 'password')
            .where({ username })
            .first()

          if (user) {
            const isPasswordValid = await fastify.compare(
              password,
              user.password
            )
            if (isPasswordValid) {
              const roles = await fastify
                .knex<{ name: string }>('roles')
                .select('roles.name')
                .join('user_roles', 'roles.id', '=', 'user_roles.role_id')
                .join('users', 'user_roles.user_id', '=', 'users.id')
                .where('users.username', username)

              request.session.user = {
                username,
                roles: roles.map((role) => role.name)
              }

              await request.session.save()

              return { success: true }
            }
          }

          reply.status(401)
          return { message: 'Invalid username or password.' }
        } catch (error) {
          await trx.rollback()
          throw error
        }
      })
    }
```

How would you test that transaction rollback on error?
I don't know if there is a way to mock/override decorators.

@climba03003

climba03003: I do see a way to monkey-patch `fastify.compare` and then throw at that point.

jean-michelet: I can do this:
```

test('Transaction should rollback on error', async (t) => {
  mock.module(
    path.join(
      import.meta.dirname,
      '../../../../src/plugins/custom/scrypt.js'
    ),
    {
      defaultExport: fastifyPlugin(
        async (fastify: FastifyInstance) => {
          fastify.decorate('compare', (value: string, hash: string) => {
            throw new Error()
          })
        },
        {
          name: 'scrypt'
        }
      )
    }
  )

  const app = await build(t)

  const res = await app.inject({
    method: 'POST',
    url: '/api/auth/login',
    payload: {
      username: 'basic',
      password: 'password123$'
    }
  })

  assert.strictEqual(res.statusCode, 500)
})
```

But it means using experimental mock.module feature and it weirdly breaks coverage...

Plz, can you give me the solution if you know it?

jean-michelet: Scrypt-coverage

jean-michelet: Ignore my previous messages, I don't know why I tough decorators were immutable...

scripts/create-database.ts
```diff
@@ -0,0 +1,26 @@
+import { createConnection, Connection } from 'mysql2/promise'
+
+async function createDatabase () {
+  const connection: Connection = await createConnection({
```
Mathieuka, line 4: Isn't the type Connection already inferred? If so, it's redundant

Mathieuka, line 4: Isn't the type Connection already inferred? If so, it's redundant

scripts/migrate.ts
```diff
@@ -0,0 +1,61 @@
+import mysql, { Connection } from 'mysql2/promise'
+import path from 'node:path'
+import fs from 'node:fs'
+import Postgrator from 'postgrator'
+
+interface PostgratorResult {
+  rows: any;
+  fields: any;
+}
```
Mathieuka, line 9:
```suggestion
interface PostgratorResult {
  rows: RowDataPacket[];
  fields: FieldPacket[];
}
```

scripts/migrate.ts
```diff
@@ -0,0 +1,61 @@
+import mysql, { Connection } from 'mysql2/promise'
```
Mathieuka, line 1:
```suggestion
import mysql, { Connection, RowDataPacket, FieldPacket } from 'mysql2/promise'
```

scripts/migrate.ts
```diff
@@ -0,0 +1,61 @@
...
+        return { rows, fields }
+      },
+      schemaTable: 'schemaversion'
+    })
+
+    await postgrator.migrate()
+
+    console.log('Migration completed!')
+
+    await new Promise<void>((resolve, reject) => {
```
Mathieuka, line 45: It doesn't seem very useful to have 2 time `connection.end`, we can include it directly in the try-catch block in the finally block ? 

```
finally {
    try {
         await connection.end();
    } catch (error) {
         console.error(error);
    }
  }
```

scripts/seed-database.ts
```diff
@@ -0,0 +1,81 @@
...
+    console.error('Error seeding database:', error)
+  } finally {
+    await connection.end()
+  }
+}
+
+async function truncateTables (connection: Connection) {
+  const [tables]: any[] = await connection.query('SHOW TABLES')
+
+  if (tables.length > 0) {
```
Mathieuka, line 27:
```suggestion
if (tables.length) {
```

scripts/seed-database.ts
```diff
@@ -0,0 +1,81 @@
...
+    const tableNames = tables.map(
+      (row: { [key: string]: string }) => row[`Tables_in_${process.env.MYSQL_DATABASE}`]
```
Mathieuka, line 29:
```suggestion
(row: Record<string, string>) => row[`Tables_in_${process.env.MYSQL_DATABASE}`]
```

scripts/seed-database.ts
```diff
@@ -0,0 +1,81 @@
...
+async function seedUsers (connection: Connection) {
+  const usernames = ['basic', 'moderator', 'admin']
+  const hash = await scryptHash('password123$')
+
+  // The goal here is to create a role hierarchy
+  // E.g. an admin should have all the roles
+  const rolesAccumulator: number[] = []
+
+  for (const username of usernames) {
+    const [userResult]: any[] = await connection.execute(`
```
Mathieuka, line 54: To improve type safety, instead of forcing the type via casting, can we create an use the `InsertResult` type ?

```
type InsertResult = {
  insertId: number;
}

const [userResult]: InsertResult[] = ...

const userId = (userResult)...
```

src/plugins/external/session.ts
```diff
@@ -0,0 +1,30 @@
+import fastifySession from '@fastify/session'
+import fp from 'fastify-plugin'
+import { Auth } from '../../schemas/auth.js'
```
Mathieuka, line 3: Path aliasing in the future?

src/routes/api/tasks/index.ts
```diff
@@ -7,25 +7,54 @@ import {
...
+      }
+
+      const tasks = await query
+        .limit(q.limit)
+        .offset(offset)
+        .orderBy('created_at', q.order)
+
+      return {
+        tasks,
+        total: tasks.length > 0 ? Number(tasks[0].total) : 0
```
Mathieuka, line 56:
```suggestion
total: tasks.length ? Number(tasks[0].total) : 0
```

scripts/migrate.ts
jean-michelet: Unfortunately this is enforced by the library I use...
```ts
  export interface BaseOptions {
    schemaTable?: string;
    validateChecksums?: boolean;
    migrationPattern?: string;
    newline?: string;
    execQuery?: (query: string) => Promise<{ rows: any[] }>;
    execSqlScript?: (sqlScript: string) => Promise<void>;
  }
```

Do you know how to extends that? Anyway, this not important for the scripts but if you know how to not use any tell me.

scripts/seed-database.ts
jean-michelet: You would perform a nullish check, although it should work here, this is semantically wrong imho.

scripts/seed-database.ts
jean-michelet: Here, `any` really is `[QueryResult, FieldPacket[]]` so it would not work, `mysql2` typing is pretty bad, they don't offer generics if I understand correctly. Maybe we could use `knex` here too. I will create an issue about it to investigate.

src/plugins/external/session.ts
jean-michelet: Unlike frontend projects, I am not sure it would work well on a node.js server.

src/routes/api/tasks/index.ts
jean-michelet: Same as my previous feedback.

jean-michelet:
I should do a thorough review, especially to make sure the doc is up to date and that we are consistent in the plugin (code in general) organization.

But I think I can get most of your feedback already so I change the status to RFR!

src/routes/api/tasks/index.ts
```diff
@@ -151,32 +187,122 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+        }
+
+        const file = await request.file()
+
+        if (!file) {
+          return reply.notFound('File not found')
+        }
+
+        const allowedMimeTypes = ['image/jpeg', 'image/png']
+        if (!allowedMimeTypes.includes(file.mimetype)) {
```
jean-michelet, line 248: I realize mime type is not guessed from file content, if I send `one_line_csv.png`, it does work...

jean-michelet: Should I use `file-type`?
```ts
const fileBuffer = await file.toBuffer();
const fileType = await fileTypeFromBuffer(fileBuffer);
```

package.json
```diff
@@ -9,29 +9,33 @@
   },
   "scripts": {
     "start": "npm run build && fastify start -l info dist/app.js",
-    "build": "tsc",
+    "build": "rm -rf ./dist && tsc",
```
Fdawgs, line 12:
```suggestion
    "build": "rm -rf ./dist && tsc",
```

This won't work for Windows users using cmd or PowerShell.
You may need to either add a script to `scripts/` that cleans up dist and call that, or explicitly state in documentation for users to use Git Bash which comes with Git for Windows.

melroy89: Or just don't use windows 🤣🤭

src/plugins/external/session.ts
jean-michelet: I have no opinion on the subject if others are ok, why not :+1:

gurgunday (approved):

climba03003 (approved):
It is a bit hard to review, as it contain nearly 50 files changes.
Overall, LGTM. Should merge it before adding more changes.

src/routes/api/tasks/index.ts
jean-michelet: No opinion about this @climba03003 @gurgunday?

jean-michelet: After this feedback, I will merge.

jean-michelet:
> It is a bit hard to review, as it contain nearly 50 files changes.

Sorry about that, I really wanted to achieve the main backend features.

src/routes/api/tasks/index.ts
Fdawgs: I'd say yes, as it's a security risk. It's something I checked in one of my APIs: https://github.com/Fdawgs/docsmith/blob/main/src/routes/pdf/html/index.js#L32-L51

jean-michelet: Great the link, this is a good reason for educating about `addContentTypeParser` utility.
But don't you think we should mention this security risk in the plugin documentation?

gurgunday: There could be a mention, yeah

But it's ultimately the same idea of never trusting a client without input validation

jean-michelet: But do we offer this kind of validation, or does a package we use offer this kind of validation?

Doesn't look like we recommend real validation for content file by reading the doc.

climba03003: I am not very strong opinion on inspecting the file content for the content-type, most of the application only use `.ext` to `content-type` (e.g. Reverse Proxy, Browser, OS).
If it is an input file (batch processing), you should validate the content inside instead of the guessed `content-type`.
If it is an image file or the others, using wrong content-type does make changes (malicious code can still be hide inside the correct content-type and execute when parsing).

metcoder95 (approved):

jean-michelet:
OK, I think I am gonna merge and will create issues to fix the existing or add new features.

It could be great to have more reviews, especially on the security aspect.

PR #46, opened 2024-10-26
feat: endpoint to delete a task image ( #37 )

### LINKED ISSUE #37: Create a endpoint to delete a task image
Currently we allow users to upload images, but not to delete them.

Turn0xx:
Fixes https://github.com/fastify/demo/issues/37

i had to make filename nullable in Task, because updating it with an undefined raised an internal error

- [x] run `npm run test` and `npm run benchmark`  -- benchmark is not disponible

### DIFF
--- a/src/routes/api/tasks/index.ts
+++ b/src/routes/api/tasks/index.ts
@@ -341,7 +341,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
         return { message: 'File deleted successfully' }
       }).catch((err) => {
         if (err.code === 'ENOENT') {
-          reply.notFound(`File "${filename}" not found`)
+          return reply.notFound(`File "${filename}" not found`)
         }
 
         reply.internalServerError('Transaction failed.')
--- a/src/routes/api/tasks/index.ts
+++ b/src/routes/api/tasks/index.ts
@@ -334,16 +334,20 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           filename
         )
 
-        await fs.promises.unlink(filePath)
+        await fs.promises.unlink(filePath).catch((err) => {
+          // Just warn if file not found as file deletion 
+          // should be treated as successful if the database operation succeeds,
+          if (err.code === 'ENOENT') {
+            fastify.log.warn(`File path '${filename}' not found`)
+          } else {
+            throw err
+          }
+        })
 
         reply.code(204)
 
         return { message: 'File deleted successfully' }
-      }).catch((err) => {
-        if (err.code === 'ENOENT') {
-          return reply.notFound(`File "${filename}" not found`)
-        }
-
+      }).catch(() => {
         reply.internalServerError('Transaction failed.')
       })
     }
--- a/src/routes/api/tasks/index.ts
+++ b/src/routes/api/tasks/index.ts
@@ -334,15 +334,17 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           filename
         )
 
-        await fs.promises.unlink(filePath).catch((err) => {
-          // Just warn if file not found as file deletion 
-          // should be treated as successful if the database operation succeeds,
-          if (err.code === 'ENOENT') {
+        try {
+          await fs.promises.unlink(filePath)
+        } catch (err) {
+          if (isErrnoException(err) && err.code === 'ENOENT') {
+            // A file could have been deleted by an external actor, e.g. system administrator.
+            // We log the error to keep a record of the failure, but consider that the operation was successful.
             fastify.log.warn(`File path '${filename}' not found`)
           } else {
             throw err
           }
-        })
+        }
 
         reply.code(204)
 
@@ -352,6 +354,10 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
       })
     }
   )
+}
+
+function isErrnoException (error: unknown): error is NodeJS.ErrnoException {
+  return error instanceof Error && 'code' in error;
 }
 
 export default plugin
--- a/src/routes/api/tasks/index.ts
+++ b/src/routes/api/tasks/index.ts
@@ -230,11 +230,6 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
       const { id } = request.params
 
       return fastify.knex.transaction(async (trx) => {
-        const task = await trx<Task>('tasks').where({ id }).first()
-        if (!task) {
-          return reply.notFound('Task not found')
-        }
-
         const file = await request.file()
         if (!file) {
           return reply.notFound('File not found')
@@ -249,7 +244,19 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           return reply.badRequest('Invalid file type')
         }
 
+
         const filename = `${id}_${file.filename}`
+
+        const numberOfRowUpdated = await trx<Task>('tasks')
+          .where({ id })
+          .update({ filename })
+
+        if (numberOfRowUpdated === 0) {
+          return reply.notFound('Task not found')
+        }
+
+
+
         const filePath = path.join(
           import.meta.dirname,
           '../../../..',
@@ -260,9 +267,6 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
 
         await pipeline(file.file, fs.createWriteStream(filePath))
 
-        await trx<Task>('tasks')
-          .where({ id })
-          .update({ filename })
 
         return { message: 'File uploaded successfully' }
       }).catch(() => {
@@ -318,24 +322,24 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
       const { filename } = request.params
 
       return fastify.knex.transaction(async (trx) => {
-        const task = await trx<Task>('tasks').select('filename').where({ filename }).first()
-        if (!task) {
+        const numberOfRowUpdated = await trx<Task>('tasks')
+          .where({ filename })
+          .update({ filename: null })
+
+        if (numberOfRowUpdated === 0) {
           return reply.notFound(`No task has filename "${filename}"`)
         }
+
 
         const filePath = path.join(
           import.meta.dirname,
           '../../../..',
           fastify.config.UPLOAD_DIRNAME,
           fastify.config.UPLOAD_TASKS_DIRNAME,
-          task.filename!
+          filename
         )
 
         await fs.promises.unlink(filePath)
-
-        await trx<Task>('tasks')
-          .where({ filename })
-          .update({ filename: null })
 
         reply.code(204).send(null)
       }).catch(() => {
--- a/src/routes/api/tasks/index.ts
+++ b/src/routes/api/tasks/index.ts
@@ -357,7 +357,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
 }
 
 function isErrnoException (error: unknown): error is NodeJS.ErrnoException {
-  return error instanceof Error && 'code' in error;
+  return error instanceof Error && 'code' in error
 }
 
 export default plugin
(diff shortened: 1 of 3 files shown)


src/routes/api/tasks/index.ts
```diff
@@ -299,6 +299,50 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+      }
+    },
+    async function (request, reply) {
+      const { filename } = request.params
+
+      return fastify.knex.transaction(async (trx) => {
+        const task = await trx<Task>('tasks').select('filename').where({ filename }).first()
+        if (!task) {
+          return reply.notFound(`No task has filename "${filename}"`)
+        }
```
jean-michelet, line 324: Maybe you can try to update, and if it return `0`, it means no row where updated, so you can guess that the task was not found. Maybe we can apply the same strategy for upload endpoint btw.

The reason is to make one query instead of 2.

test/routes/api/tasks/tasks.test.ts
```diff
@@ -432,7 +432,7 @@ describe('Tasks api (logged user only)', () => {
     })
   })
 
-  describe('Task image upload and retrieval', () => {
+  describe('Task image upload and retrieval and delete', () => {
```
jean-michelet, line 435: You can keep the title I think or create a more generic title.
Adding `and` is the sign that we are doing too much things in the same place.
Maybe we should create 3 (child?) `describe` that share some utilities.

jean-michelet (commented):
You must reach 100% code coverage, meaning all the code must be executed during test or annotated with `/* c8 ignore` for very good reasons.

jean-michelet:
Edit the description and add a link the issue plz, it allows people following the issue to be notified and readers to be aware that a PR has been pushed.

src/routes/api/tasks/index.ts
Turn0xx: Ok, should i make the change for the upload endpoint also or i can do it in another pr / issue ?

jean-michelet: Yep, I think you can do it on this PR if you want.
You put the logic after file size/mime validation.

test/routes/api/tasks/tasks.test.ts
Turn0xx: Actually, all tests in the task image bloc share the same state, because in retrival & deletion tests, we act on the image uploaded in the first upload test ! what do you think about just making a generic title and splitting test in describe child ?

jean-michelet: > all tests in the task image bloc share the same state
I know, if achievable with child describe, it's fine.

Turn0xx:
@jean-michelet i think i break something in tests (they all pass but there is some "unsettled top-level await") at the end of the test suites

test/routes/api/tasks/tasks.test.ts
```diff
@@ -457,182 +471,277 @@ describe('Tasks api (logged user only)', () => {
...
-      assert.strictEqual(res.headers['content-type'], 'image/png')
+      beforeEach(async () => {
+        app = await build()
+        await uploadImageForTask(app, taskId, testImagePath, uploadDirTask)
+        await app.close()
+      })
 
-      const originalFile = fs.readFileSync(testImagePath)
+      after(async () => {
+        app = await build()
```
jean-michelet, line 682: Why do you build the app here? 
Here you are building to app but only closing one, maybe this is the reason of you bug.
If you don't pass `t` (TestContext) to build, it will not close automatically, this is why we need to close it in `after` hook.

src/routes/api/tasks/index.ts
```diff
@@ -249,7 +244,19 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           return reply.badRequest('Invalid file type')
         }
 
+
         const filename = `${id}_${file.filename}`
+
+        const numberOfRowUpdated = await trx<Task>('tasks')
```
jean-michelet, line 250: Maybe you can shorter the name e.g.  `affectedRows`.

src/routes/api/tasks/index.ts
```diff
@@ -299,6 +303,50 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+          import.meta.dirname,
+          '../../../..',
+          fastify.config.UPLOAD_DIRNAME,
+          fastify.config.UPLOAD_TASKS_DIRNAME,
+          filename
+        )
+
+        await fs.promises.unlink(filePath)
+
+        reply.code(204).send(null)
```
jean-michelet, line 344: You can also return a more friendly response message:
```js
{ message: 'File uploaded successfully' }
```

test/routes/api/tasks/tasks.test.ts
```diff
@@ -457,182 +471,277 @@ describe('Tasks api (logged user only)', () => {
...
+      })
+
+      it('File deletion transaction should rollback on error', async (t) => {
+        const app = await build(t)
+        const { mock: mockPipeline } = t.mock.method(fs.promises, 'unlink')
+        mockPipeline.mockImplementationOnce(() => {
+          throw new Error()
+        })
+
+        const { mock: mockLogError } = t.mock.method(app.log, 'error')
```
jean-michelet, line 727: I think you can just save a task with a filename that does not exist in upload directory and it will trigger a `ENOENT` error, ensuring the transaction rollback.

src/routes/api/tasks/index.ts
```diff
@@ -299,6 +303,50 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+
+        const filePath = path.join(
+          import.meta.dirname,
+          '../../../..',
+          fastify.config.UPLOAD_DIRNAME,
+          fastify.config.UPLOAD_TASKS_DIRNAME,
+          filename
+        )
+
+        await fs.promises.unlink(filePath)
```
jean-michelet, line 342: If `ENOENT` error, we should return 404 Not found in the `catch` statement.

test/routes/api/tasks/tasks.test.ts
jean-michelet: https://github.com/fastify/demo/pull/46#discussion_r1818070252

jean-michelet: You'll have to throw an other error than `ENOENT`, so you can keep the mock.

src/routes/api/tasks/index.ts
Turn0xx: Then i have to mock the unlink function for the rollback test, no ?

jean-michelet: Not for 404 error if file is not found. But yes for throwing an error that will lead to 500 internal server error.

test/routes/api/tasks/tasks.test.ts
Turn0xx: i have a really weird behavior with this bug, i just deleted the build and the close in the after hook, make some changes for other comments, all works fine, the bug disappeared, then i added a another tests for 100% coverage (the 404 file note found for the unlink) and then the bug appeared again.

jean-michelet: When yo uremove test `should return 404 for non-existant file in upload dir`, it does work?

Turn0xx: Yes exactly !

src/routes/api/tasks/index.ts
```diff
@@ -299,6 +299,55 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+        reply.code(204)
+
+        return { message: 'File deleted successfully' }
+      }).catch((err) => {
+        if (err.code === 'ENOENT') {
+          reply.notFound(`File "${filename}" not found`)
+        }
+
+        reply.internalServerError('Transaction failed.')
+      })
```
jean-michelet, line 348: Add `return` statements plz.

test/routes/api/tasks/tasks.test.ts
```diff
@@ -432,7 +444,7 @@ describe('Tasks api (logged user only)', () => {
     })
   })
 
-  describe('Task image upload and retrieval', () => {
+  describe('Task image upload, retrieval and delete', (context) => {
```
jean-michelet, line 447:
```suggestion
  describe('Task image upload, retrieval and delete', () => {
```

test/routes/api/tasks/tasks.test.ts
```diff
@@ -457,182 +469,291 @@ describe('Tasks api (logged user only)', () => {
...
+        const app = await build(t)
+
+        await createTask(app, {
+          name: 'Task with image',
+          author_id: 1,
+          status: TaskStatusEnum.New,
+          filename: 'does_not_exist.png'
+        })
+
+        const resDelete = await app.injectWithLogin('basic', {
```
jean-michelet, line 723:
```suggestion
        const res = await app.injectWithLogin('basic', {
```

`res` is enough context to understand the code.

test/routes/api/tasks/tasks.test.ts
```diff
@@ -457,182 +469,291 @@ describe('Tasks api (logged user only)', () => {
...
+        const app = await build(t)
+        const { mock: mockPipeline } = t.mock.method(fs.promises, 'unlink')
+        mockPipeline.mockImplementationOnce(() => {
+          return Promise.reject(new Error())
+        })
+
+        const { mock: mockLogError } = t.mock.method(app.log, 'error')
+
+        const taskFilename = encodeURIComponent(`${taskId}_${filename}`)
+        const resDelete = await app.injectWithLogin('basic', {
```
jean-michelet, line 743: Same.

jean-michelet (approved):
Look at my last feedbacks and it will be ok for me, good job :clap:

src/routes/api/tasks/index.ts
climba03003: I always prefer consider it as success even no file is deleted.
The reason is very simple, file system can be manged outside of the system. It is easy to have mismatch in file system and database when the system admin did something wrong. I consider when database operation is success, it is always success (no matter the result of file system operation).

This only valid for removing file, update or create should requires both operation to success.

Turn0xx: So i should catch the unlink error and just send a 204 response ?

climba03003: That's only my personal preference, raised up for reference only.
Not necessary to change.

jean-michelet: Yep, good point.


Maybe we can still log the error in case of a real issue?
```ts
await fs.promises.unlink('filePath').catch(() => {
  fastify.log.warn(`File path '${filePath}' not found`)
});
```

jean-michelet: > That's only my personal preference,

I think this is objectively the right approach.

jean-michelet: Can you do the change @Turn0xx, then I merge.

Turn0xx: Yes, i gonna push this morning 👌

jean-michelet: I would add a comment that sums up Climba's point, this demo is meant to be pedagogic.

src/routes/api/tasks/index.ts
```diff
@@ -334,16 +334,20 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           filename
         )
 
-        await fs.promises.unlink(filePath)
+        await fs.promises.unlink(filePath).catch((err) => {
+          // Just warn if file not found as file deletion 
+          // should be treated as successful if the database operation succeeds,
```
jean-michelet, line 339:
```suggestion
          // A file could have been deleted by an external actor, e.g. system administrator.
          // We log the error to keep a record of the failure, but consider that the operation was successful.
```

Use directly a try/catch instead of `.catch`, I was wrong to offer to handle this way.

jean-michelet (changes requested):

PR #48, opened 2024-10-28
Add auth update endpoint

### LINKED ISSUE #36: Create endpoints to allow users to update their profile information and passwords.

Mathieuka:
### Description

Fixes https://github.com/fastify/demo/issues/36

Key Changes

-  Allows authenticated users to change their passwords using a PUT request with schema validation.
- Enforces a strong password policy requiring uppercase, lowercase, numeric, and special characters.
- Rate Limiting: Limits password update attempts to 3 per minute to enhance security.
- Includes tests for successful updates, validation errors, and rate limiting.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`

### DIFF
--- a/src/routes/api/auth/index.ts
+++ b/src/routes/api/auth/index.ts
@@ -2,7 +2,7 @@ import {
   FastifyPluginAsyncTypebox,
   Type
 } from '@fastify/type-provider-typebox'
-import { CredentialsSchema, Credentials } from '../../../schemas/auth.js'
+import { CredentialsSchema, UpdateCredentialsSchema, Credentials, Auth } from '../../../schemas/auth.js'
 
 const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
   fastify.post(
@@ -60,6 +60,48 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
       }).catch(() => {
         reply.internalServerError('Transaction failed.')
       })
+    }
+  )
+  fastify.post(
+    '/update',
+    {
+      schema: {
+        body: UpdateCredentialsSchema,
+        response: {
+          200: Type.Object({
+            message: Type.String()
+          }),
+          401: Type.Object({
+            message: Type.String()
+          })
+        },
+        tags: ['Authentication']
+      }
+    },
+    async function (request, reply) {
+      const { username } = request.body
+      const user = await fastify.repository.find<Auth>('users', {
+        select: 'username, password',
+        where: { username }
+      })
+
+      if (user) {
+        const isPasswordValid = await fastify.compare(request.body.password, user.password)
+
+        if (isPasswordValid) {
+          const hashedPassword = await fastify.hash(request.body.newPassword || request.body.password)
+          await fastify.repository.update('users', {
+            data: { password: hashedPassword, username: request.body.newUsername || request.body.username },
+            where: { username }
+          })
+
+          return { message: 'Password updated successfully' }
+        }
+      }
+
+      reply.status(401)
+
+      return { message: 'Invalid username or password.' }
     }
   )
 }
--- a/src/routes/api/auth/index.ts
+++ b/src/routes/api/auth/index.ts
@@ -2,7 +2,7 @@ import {
   FastifyPluginAsyncTypebox,
   Type
 } from '@fastify/type-provider-typebox'
-import { CredentialsSchema, UpdateCredentialsSchema, Credentials, Auth } from '../../../schemas/auth.js'
+import { CredentialsSchema, Credentials } from '../../../schemas/auth.js'
 
 const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
   fastify.post(
@@ -36,6 +36,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
             password,
             user.password
           )
+
           if (isPasswordValid) {
             const roles = await trx<{ name: string }>('roles')
               .select('roles.name')
@@ -55,53 +56,10 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
         }
 
         reply.status(401)
-
         return { message: 'Invalid username or password.' }
       }).catch(() => {
         reply.internalServerError('Transaction failed.')
       })
-    }
-  )
-  fastify.post(
-    '/update',
-    {
-      schema: {
-        body: UpdateCredentialsSchema,
-        response: {
-          200: Type.Object({
-            message: Type.String()
-          }),
-          401: Type.Object({
-            message: Type.String()
-          })
-        },
-        tags: ['Authentication']
-      }
-    },
-    async function (request, reply) {
-      const { username } = request.body
-      const user = await fastify.repository.find<Auth>('users', {
-        select: 'username, password',
-        where: { username }
-      })
-
-      if (user) {
-        const isPasswordValid = await fastify.compare(request.body.password, user.password)
-
-        if (isPasswordValid) {
-          const hashedPassword = await fastify.hash(request.body.newPassword || request.body.password)
-          await fastify.repository.update('users', {
-            data: { password: hashedPassword, username: request.body.newUsername || request.body.username },
-            where: { username }
-          })
-
-          return { message: 'Password updated successfully' }
-        }
-      }
-
-      reply.status(401)
-
-      return { message: 'Invalid username or password.' }
     }
   )
 }
--- a/src/routes/api/auth/index.ts
+++ b/src/routes/api/auth/index.ts
@@ -36,7 +36,6 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
             password,
             user.password
           )
-
           if (isPasswordValid) {
             const roles = await trx<{ name: string }>('roles')
               .select('roles.name')
@@ -56,6 +55,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
         }
 
         reply.status(401)
+
         return { message: 'Invalid username or password.' }
       }).catch(() => {
         reply.internalServerError('Transaction failed.')
--- a/src/routes/api/user/index.ts
+++ b/src/routes/api/user/index.ts
@@ -46,25 +46,21 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           .select('username', 'password')
           .where({ username })
           .first()
+        const isPasswordValid = await fastify.compare(currentPassword, user.password)
 
-        if (user) {
-          const isPasswordValid = await fastify.compare(currentPassword, user.password)
-
-          if (isPasswordValid) {
-            const hashedPassword = await fastify.hash(newPassword)
-
-            await fastify.knex('users')
-              .update({
-                password: hashedPassword
-              })
-              .where({ username })
-
-            return { message: 'Password updated successfully' }
-          }
+        if (!user || !isPasswordValid) {
+          reply.status(401)
+          return { message: 'Invalid username or password.' }
         }
 
-        reply.status(401)
-        return { message: 'Invalid username or password.' }
+        const hashedPassword = await fastify.hash(newPassword)
+        await fastify.knex('users')
+          .update({
+            password: hashedPassword
+          })
+          .where({ username })
+
+        return { message: 'Password updated successfully' }
       } catch (error) {
         reply.internalServerError('An error occurred while updating the password.')
       }
--- a/src/routes/api/user/index.ts
+++ b/src/routes/api/user/index.ts
@@ -54,9 +54,8 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           .first()
         const isPasswordValid = await fastify.compare(currentPassword, user.password)
 
-        if (!user || !isPasswordValid) {
-          reply.status(401)
-          return { message: 'Invalid username or password.' }
+        if (!user) {
+          return reply.unauthorized()
         }
 
         const hashedPassword = await fastify.hash(newPassword)
--- a/src/routes/api/user/index.ts
+++ b/src/routes/api/user/index.ts
@@ -39,6 +39,12 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
     },
     async function (request, reply) {
       const { newPassword, currentPassword } = request.body
+
+      if (newPassword === currentPassword) {
+        reply.status(400)
+        return { message: 'New password cannot be the same as the current password.' }
+      }
+
       const username = request.session.user.username
 
       try {
--- a/src/routes/api/user/index.ts
+++ b/src/routes/api/user/index.ts
@@ -48,7 +48,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           .first()
 
         if (!user) {
-          return reply.unauthorized()
+          return reply.code(401).send({ message: 'User does not exist.' })
         }
 
         const isPasswordValid = await fastify.compare(
--- a/src/routes/api/user/index.ts
+++ b/src/routes/api/user/index.ts
@@ -19,7 +19,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
             message: Type.String()
           })
         },
-        tags: ['Users']
+        tags: ['User']
       }
     },
     async function (request, reply) {
--- a/src/routes/api/user/index.ts
+++ b/src/routes/api/user/index.ts
@@ -39,12 +39,6 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
     },
     async function (request, reply) {
       const { newPassword, currentPassword } = request.body
-
-      if (newPassword === currentPassword) {
-        reply.status(400)
-        return { message: 'New password cannot be the same as the current password.' }
-      }
-
       const username = request.session.user.username
 
       try {
@@ -52,10 +46,14 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           .select('username', 'password')
           .where({ username })
           .first()
-        const isPasswordValid = await fastify.compare(currentPassword, user.password)
 
         if (!user) {
           return reply.unauthorized()
+        }
+
+        if (newPassword === currentPassword) {
+          reply.status(400)
+          return { message: 'New password cannot be the same as the current password.' }
         }
 
         const hashedPassword = await fastify.hash(newPassword)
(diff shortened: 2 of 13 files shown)


src/routes/api/auth/index.ts
```diff
@@ -43,6 +42,47 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
       return { message: 'Invalid username or password.' }
     }
   )
-}
+  fastify.post(
+    '/update',
```
climba03003, line 46: This endpoint is strange, `update` should be expressed by `method`.

src/routes/api/auth/index.ts
```diff
@@ -43,6 +42,47 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+        where: { username }
+      })
+
+      if (user) {
+        const isPasswordValid = await fastify.compare(request.body.password, user.password)
 
+        if (isPasswordValid) {
+          const hashedPassword = await fastify.hash(request.body.newPassword || request.body.password)
+          await fastify.repository.update('users', {
+            data: { password: hashedPassword, username: request.body.newUsername || request.body.username },
```
climba03003, line 74: I don't see it is a good idea to update `username` without database unique constraint or transaction checking on data.

climba03003 (changes requested):

src/routes/api/auth/index.ts
jean-michelet: Yep, we should use `PUT` if we allow to update the resource entirely or `PATCH` if we apply partial changes.

src/schemas/auth.ts
```diff
@@ -5,4 +5,11 @@ export const CredentialsSchema = Type.Object({
   password: Type.String()
 })
 
+export const UpdateCredentialsSchema = Type.Object({
+  username: Type.String(),
+  password: Type.String(),
+  newUsername: Type.Optional(Type.String()),
+  newPassword: Type.Optional(Type.String())
+})
```
jean-michelet, line 13:
```ts
export const UpdateCredentialsSchema = Type.Intersect([
  CredentialsSchema,
  Type.Object({
    newPassword: Type.String(),
  }),
]);
```

`newUsername` is not needed.

jean-michelet (commented):
Thanks for the PR!

Always add a link to the issue when you push a PR, e.g. Fixes #36 
It allows people following the issue to be notified and readers of the issue to know that a PR has been pushed.

Your branch is out of date, you need to sync your fork with this repo, pull the changes and resolve conflicts.

`/auth` is related to authentication, it must be used for login, logout, reset password, etc.

Regarding the CRUD of user details, you need to create a new folder `routes/api/users` (the [autoloader](https://github.com/fastify/fastify-autoload) will automatically prefix all the routes defined in this folder with path `/api/users`). 

I wouldn't mix password update with other user details update, as password update require specific logic and is more security sensible. Most of the time, updating user details only require payload validation.

src/schemas/auth.ts
Mathieuka: > ```ts
> export const UpdateCredentialsSchema = Type.Intersect([
>   CredentialsSchema,
>   Type.Object({
>     newPassword: Type.String(),
>   }),
> ]);
> ```
> 
> `newUsername` is not needed.

I also added `currentPassword`

src/routes/api/user/index.ts
```diff
@@ -0,0 +1,60 @@
...
+      const { newPassword, currentPassword } = request.body
+      const username = request.session.user.username
+
+      try {
+        const user = await fastify.knex<Auth>('users')
+          .select('username', 'password')
+          .where({ username })
+          .first()
+
+        if (user) {
```
jean-michelet, line 35:
```suggestion
        if (!user) {
```
Prefer early return, if there is no user, the program can stop.

src/routes/api/user/index.ts
```diff
@@ -0,0 +1,60 @@
...
+        },
+        tags: ['User']
+      }
+    },
+    async function (request, reply) {
+      const { newPassword, currentPassword } = request.body
+      const username = request.session.user.username
+
+      try {
+        const user = await fastify.knex<Auth>('users')
```
jean-michelet, line 30: All these operations should be wrapped in a transaction to avoid race conditions.

Mathieuka:
This is my first PR, I'm waiting for your review before committing to writing a test

src/schemas/users.ts
```diff
@@ -0,0 +1,6 @@
+import { Type } from '@sinclair/typebox'
+
+export const UpdateCredentialsSchema = Type.Object({
+  currentPassword: Type.String(),
+  newPassword: Type.String()
```
jean-michelet, line 5: I would force to use a strong password and check that the two values are different.

src/routes/api/user/index.ts
```diff
@@ -0,0 +1,60 @@
+import {
+  FastifyPluginAsyncTypebox,
+  Type
+} from '@fastify/type-provider-typebox'
+import { Auth } from '../../../schemas/auth.js'
+import { UpdateCredentialsSchema } from '../../../schemas/users.js'
+
+const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
+  fastify.put(
```
jean-michelet, line 9: I think we should use a rate limiter to mitigate brute force attack, there is an example here: https://github.com/fastify/demo/blob/4470023655ea4e68de65d9b4e78a6f03c56ab49a/src/app.ts#L74

src/routes/api/user/index.ts
```diff
@@ -0,0 +1,60 @@
...
+                password: hashedPassword
+              })
+              .where({ username })
+
+            return { message: 'Password updated successfully' }
+          }
+        }
+
+        reply.status(401)
+        return { message: 'Invalid username or password.' }
```
jean-michelet, line 52: This error message is wrong.

jean-michelet (commented):
You need to write tests that prove that the feature work as expected and reach 100% code coverage: https://en.wikipedia.org/wiki/Code_coverage

jean-michelet:
> This is my first PR, I'm waiting for your review before committing to writing a test

Congrats for your first PR :+1:  
Keep in mind that writing test does help you to produce robust and relevant code and save code review time.

Mathieuka:
> > This is my first PR, I'm waiting for your review before committing to writing a test
> 
> Congrats for your first PR :+1:  
> Keep in mind that writing tests does help you to produce robust and relevant code and then save code review time.


Yes 🙂, I had left a comment about the test, I prefer a double check before testing a bad design for my first contributions

https://github.com/fastify/demo/pull/48#issuecomment-2445262719

src/schemas/users.ts
```diff
@@ -0,0 +1,6 @@
...
+})
```
climba03003, line 6:
```suggestion
const Password = Type.String({
  // at least 1 upper case, 1 lower case, 1 numeric, 1 special characters with minimal 8 characters.
  pattern: "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{8,}$",
  minLength: 8
})

export const UpdateCredentialsSchema = Type.Object({
  currentPassword: Password,
  newPassword: Password
})
```

src/routes/api/user/index.ts
Mathieuka: > This error message is wrong.
I think we should not return to the client if the user or password is wrong, it is not a good security design to prevent potential attackers from getting information about the part of the login information that is incorrect.

Or i have missed something

src/schemas/users.ts
Mathieuka: > const Password = Type.String({
>   // at least 1 upper case, 1 lower case, 1 numeric, 1 special characters with minimal 8 characters.
>   pattern: "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).{8,}$",
>   minLength: 8
> })

Thanks for the suggestion

The minimum length rule is duplicated in both the regex pattern and the JSON Schema definition, i'll the rules from the schema

src/routes/api/user/index.ts
jean-michelet: > I think we should not return to the client if the user or password is wrong

For a auth endpoint yes, for a setting endpoint, yes, because we know this is the user/admin that do this operation (unless hacked, this is why we have rate limiter). We must inform why the request for password update failed.

Regarding the line `52`, if there is no user session, the error should be detected ahead of time in this hook:
https://github.com/fastify/demo/blob/4470023655ea4e68de65d9b4e78a6f03c56ab49a/src/routes/api/autohooks.ts#L9

But you can still return: ` reply.unauthorized()`, it will set status code to `401` automatically.

src/schemas/users.ts
climba03003: > the pattern defined only match strings that are exactly 4 characters long

No, it matches any string that is 8 characters or longer. See the last part `.{8,}`.
image

> the minimum length rule is duplicated in both the regex pattern and the JSON Schema definition, i'll keep the rules from the schema.

By looking at the pattern, it is very hard to understand how it works and does it limit the length.
The `minLength` is used here for ensure the string is at least 8 characters and faster understanding on it.
I don't knows which one take higher priority to check in `ajv` but more details is better.

Mathieuka: > "The minLength is used here for ensure the string is at least 8 characters and faster understanding on it."

True, thanks for your answer

test/routes/api/user/user.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      method: 'PUT',
+      url: '/api/user/update-password',
+      payload: {
+        currentPassword: 'Password123$',
+        newPassword: 'password123$'
+      }
+    })
+
+    assert.strictEqual(res.statusCode, 400)
+    assert.deepStrictEqual(JSON.parse(res.payload), { message: 'body/newPassword must match pattern "^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).*$"' })
```
Mathieuka, line 114: I should find a way to return a more explicit error message

jean-michelet:
Code coverage is not reached, meaning some part of the code was not executed during test.

test/routes/api/user/user.test.ts
jean-michelet: > Password must contain at least one uppercase letter, one lowercase letter, one digit, and one special character (#?!@$%^&*-).

test/routes/api/user/user.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  before(async () => {
+    const app = await build()
+    const users = ['random-user-0', 'random-user-1', 'random-user-2', 'random-user-3', 'random-user-4', 'random-user-5']
+
+    for (const user of users) {
+      await createUser(app, {
+        username: user,
+        password: Password123$
+      })
+    }
```
jean-michelet, line 37: I don't understand why you create all the users here, than recreate them individually in all the tests.

test/routes/api/user/user.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    await createUser(app, { username: 'random-user-0', password: Password123$ })
+    const res = await updatePasswordWithLoginInjection(app, 'random-user-0', {
+      currentPassword: 'Password123$',
+      newPassword: 'NewPassword123$'
+    })
+
+    assert.strictEqual(res.statusCode, 200)
+    assert.deepStrictEqual(JSON.parse(res.payload), { message: 'Password updated successfully' })
+
+    await deleteUser(app, 'random-user-0')
```
jean-michelet, line 60: Why do you delete the users at the end of each test?
Just clean all of them in `after` hook when you are done.

test/routes/api/user/user.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      return await updatePasswordWithLoginInjection(app, 'random-user-5', {
+        currentPassword: 'WrongPassword123$',
+        newPassword: 'Password123$'
+      })
+    }
+
+    await updatePassword()
+    await updatePassword()
+    await updatePassword()
+    const res = await updatePassword()
```
jean-michelet, line 115: You can also do a simple for loop.

src/routes/api/user/index.ts
```diff
@@ -0,0 +1,83 @@
...
+          max: 3,
+          timeWindow: '1 minute',
+          errorResponseBuilder: function (_, context) {
+            return {
+              statusCode: 429,
+              error: 'Too Many Requests',
+              message: `You have reached the request limit. Please try again in ${Math.floor(context.ttl / 1000)} seconds.`,
+              date: new Date().toISOString(),
+              retryAfter: context.ttl
+            }
```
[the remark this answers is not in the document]
Mathieuka: No need to override the built in error message, I'll delete it

test/routes/api/user/user.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  return await app.injectWithLogin(username, {
+    method: 'PUT',
+    url: '/api/user/update-password',
+    payload
+  })
+}
+
+describe('User API', () => {
+  // Hashed value of `Password123$`
+  const Password123$ = 'ff57faf149a2bcab41bf7ecbbc8ce491.3ce6b34ea3edb3f0a09f811440885bfeda612832c04bfddc9d4b906019d97fa0'
```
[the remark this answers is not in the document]
Mathieuka: Good point

src/routes/api/user/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      try {
+        const user = await fastify.knex<Auth>('users')
+          .select('username', 'password')
+          .where({ username })
+          .first()
+
+        if (!user) {
+          return reply.code(401).send({ message: 'User does not exist.' })
+        }
```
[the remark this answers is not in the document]
Mathieuka: Isn't it possible for a user to be deleted between the time they authenticate and the time they try to change their password?

By themselves or an admin

99.999% -> ?%

test/routes/api/user/user.test.ts
Mathieuka: it's an oversight

src/routes/api/user/index.ts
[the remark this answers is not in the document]
jean-michelet: > Isn't it possible for a user to be deleted between the time they authenticate and the time they try to change their password?

There is no user administration for now, but this is a valid point. I think we should be able to invalidate/update a session when we remove a user or change any property that is related to authentication like `roles`.
Maybe we can use Redis to store session with user id so we can write to it.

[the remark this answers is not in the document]
jean-michelet: > Maybe we can use Redis to store session

I will create an issue about this.

test/routes/api/user/user.test.ts
Mathieuka: Tests should be run in isolation as much as possible, shared dependencies across preparation hooks are hell.

But if that's a style you prefer, I can use hooks

jean-michelet: I told you that because you were creating all the users in the `before` hook, but I agree with you.

src/schemas/users.ts
```diff
@@ -0,0 +1,13 @@
+import { Type } from '@sinclair/typebox'
+
+const passwordPattern = '^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).*$'
+
+const PasswordSchema = Type.String({
+  pattern: passwordPattern,
```
Mathieuka, line 6: @climba03003 @jean-michelet 

Currently I don't think there is a way to customize the error message in case the payload doesn't match the pattern via the directly via the Type Builder Typebox.

I managed to do this via implementing fastify.setErrorHandler in the password update endpoint configuration, have you ever encountered this problem ?

Is it correct to manage the error message in a custom way at the endpoint ?

src/routes/api/user/index.ts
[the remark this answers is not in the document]
climba03003: > How do you handle that?

I use constraint to derive the route handler before, but currently I handle it inside one handler.
The user information is retrieved way before the handler and I would use it as reference. 
Something like `requset.user` if you have logged in.

src/schemas/users.ts
climba03003: > Is it correct to manage the error message in a custom way at the endpoint ?

You can using `ajv-errors`, so the message can be customize per schema.
https://www.npmjs.com/package/ajv-errors

climba03003 (approved):

src/schemas/users.ts
Mathieuka: @jean-michelet 

Regarding the custom error message for AJV formatting, I haven't found a solution, and it seems that @climba03003  hasn't either ? as they approved the PR following my message.

https://github.com/fastify/demo/pull/48#discussion_r1835433228

As for SonarQube's warning about duplication, that's code I don't want to abstract and SonarQube doesn't know the difference

jean-michelet:
In RESTful architecture, resource naming conventions recommend using plural nouns for resource names.
Ple, change `user` for `users`, with `s`.

src/schemas/users.ts
climba03003: We can review on the dashboard and claim it is fixed or safe.

jean-michelet: @Eomm 
Is it possible to give these rights to the Fastify members that collaborates on this repo? 
Besides me, I am aware of @climba03003 and @Fdawgs.

climba03003: I believe it can identify the Github account and give the correct permission?
Because I have updated the sonarcloud issue when I using the Github OAuth.

src/routes/home.ts
```diff
@@ -7,6 +7,12 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
   fastify.get(
     '/',
     {
+      config: {
```
Mathieuka, line 10: @jean-michelet 

I have changed the configuration of the rate limit in `home` because the environment variable `RATE_LIMIT_MAX` was recently introduced with a value of 10 by default, certainly for one of the Tasks tests which fails if we set it to 3.

3 is the default value for the rate limit at the plugin level

src/routes/api/users/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      const username = request.session.user.username
+
+      try {
+        const user = await fastify.knex<Auth>('users')
+          .select('username', 'password')
+          .where({ username })
+          .first()
+
+        if (!user) {
+          return reply.code(401).send({ message: 'User does not exist.' })
```
lhousaine, line 42: To enhance more security, I prefer giving less information to the end user by returning a general error message for both username and password. ex: username and password combination is not valid.

mcollina (approved):
lgtm

Issue #38, opened 2024-10-21
Check for existing file when uploading image


jean-michelet:
When uploading a task image, we should look if there is already a file associated, and remove it.

JP-sDEV:
Hi can I be assigned this issue?

jean-michelet:
Hi!

Go ahead!

jean-michelet:
This is `POST /:id/upload`.
We store the filename in the `tasks` table in db for later retrieval.
Look how we do store the file and you will know how to check if there is already an existing file.
When we retrieve a task, we must first look if a file with  name = `task.filename` exists in the upload directory.
But you should delete the file it after the db operations.

JP-sDEV:
Understood, I believe I have an idea on where to put the deletion. I'll open a PR hopefully tomorrow morning/afternoon

PR #50, opened 2024-10-29
fix: delete file if exist before uploading another one

Turn0xx:
-I couldn't do this with one query because MySql doesn't support RETURNING clause
-For retriveal tests I had to build a clean state for them to work because they were depending on Upload tests

Issue: #38 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`


src/routes/api/tasks/index.ts
```diff
@@ -360,4 +343,30 @@ function isErrnoException (error: unknown): error is NodeJS.ErrnoException {
...
+  } catch (err) {
+    if (isErrnoException(err) && err.code === 'ENOENT') {
+      // A file could have been deleted by an external actor, e.g. system administrator.
+      // We log the error to keep a record of the failure, but consider that the operation was successful.
+      fastify.log.warn(`File path '${fileName}' not found`)
+    } else {
+      throw err
+    }
+  }
+}
```
jean-michelet, line 370: I think we should create a new custom plugin for file operations .

We could use it a similar way:
```ts
fastify.fileHandler.uploadTaskImage(filename)
fastify.fileHandler.removeTaskImage(filename)

// Later
fastify.fileHandler.uploadUserImage(filename)
fastify.fileHandler.removeUserImage(filename)
```

Or:
```ts
fastify.fileHandler.upload(taskDirName, filename)
fastify.fileHandler.remove(taskDirName, filename)
```

src/plugins/custom/file-handling.ts
```diff
@@ -0,0 +1,76 @@
...
+import { FastifyInstance } from 'fastify'
+import path from 'path'
+import { pipeline } from 'node:stream/promises'
+import fs from 'node:fs'
+
+declare module 'fastify' {
+  export interface FastifyInstance {
+    fileHandler: {
+      upload: (fileDir: FileDir, fileName: string, file: BusboyFileStream) => Promise<void>
+      remove: (fileDir: FileDir, fileName: string) => Promise<void>
```
jean-michelet, line 12:
```suggestion
      upload: typeof upload;
      remove: typeof remove;
```

src/plugins/custom/file-handling.ts
```diff
@@ -0,0 +1,76 @@
...
+  [K in keyof T]: K extends `UPLOAD_${string}_DIRNAME` ? K : never
+}[keyof T]
+
+type FileDirMap = {
+  [key in FileDir]: FileDirConfigPath<FileDirConfig>
+}
+
+const fileDirMap: FileDirMap = {
+  tasks: 'UPLOAD_TASKS_DIRNAME'
+}
```
jean-michelet, line 44: This is unnecessary complex, just pass the path as a string argument.

src/plugins/custom/file-handling.ts
```diff
@@ -0,0 +1,76 @@
...
+
+function isErrnoException (error: unknown): error is NodeJS.ErrnoException {
+  return error instanceof Error && 'code' in error
+}
+
+export default fp(async (fastify) => {
+  fastify.decorate('fileHandler', {
+    upload: upload.bind(fastify),
+    remove: remove.bind(fastify)
```
jean-michelet, line 74: This is not needed. 

Fastify instance (`this`) will be automatically passed if we use function declaration with `function` keyword (doesn't work with arrow functions).

src/plugins/custom/file-handling.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+}
+
+interface FileDirConfig {
+  UPLOAD_TASKS_DIRNAME: string
+}
+
+function buildFilePath (fileDir: FileDir, filename: string, fastify: FastifyInstance) {
```
jean-michelet, line 22:
```suggestion
function buildFilePath (fastify: FastifyInstance, fileDir: FileDir, filename: string) {
```

test/routes/api/tasks/tasks.test.ts
```diff
@@ -622,6 +647,34 @@ describe('Tasks api (logged user only)', () => {
...
+      before(async () => {
+        app = await build()
+
+        await app.knex<Task>('tasks').where({ id: taskId }).update({ filename: null })
+
+        const files = fs.readdirSync(uploadDirTask)
+        files.forEach((file) => {
+          const filePath = path.join(uploadDirTask, file)
+          fs.rmSync(filePath, { recursive: true })
+        })
```
jean-michelet, line 659: You can create a function `removeFiles`, but the tests are run sequentially for db consistency, if call it in after hook, they will be removed before you reach this test suite.

src/plugins/custom/file-handling.ts
Turn0xx: i can't do this, because i have to bind manually the fastify instance in the decorate function

jean-michelet: > i have to bind manually the fastify instance in the decorate function

Why?

src/plugins/custom/file-handling.ts
Turn0xx: I have to bind it manually, otherwise the instance passed is incorrect.. (empty object) 
I think that it is because of the way i use the `decorate` function, I passe to it an objects and not a function, so the functions inside of the object doesn't get the right instance.

test/routes/api/tasks/tasks.test.ts
Turn0xx: > but the tests are run sequentially for db consistency, if called in after hook, they will be removed before you reach this test suite.

i didn't understand this

src/plugins/custom/file-handling.ts
jean-michelet: > I passe to it an objects and not a function

Ha, ok I reviewed too quickly.

You can put all the functions you use directly in the registration callback so you get access to the fastify instance.
```ts
export default fp(async (fastify) => {
// do all the logic here
```

test/routes/api/tasks/tasks.test.ts
jean-michelet: I mean that tests are not run in parralel, if you remove the files you uploaded during a test suite in `after` hook, it hsoudl be enough. You should not have to remove the files in `before` hook.

src/plugins/custom/file-handling.ts
Turn0xx: It's works, but i can't do this (upload and remove not anymore in scope)
```typescript
declare module 'fastify' {
  export interface FastifyInstance {
    fileHandler: {
      upload: typeof upload
      remove: typeof remove
    }
  }
}
```

jean-michelet:
Install and run Sonarqube locally before pushing. They are also editor
extensions.

src/plugins/custom/file-handling.ts
```diff
@@ -0,0 +1,59 @@
...
+  return path.join(
+    import.meta.dirname,
+    '../../../',
+    fastify.config.UPLOAD_DIRNAME,
+    fileDir,
+    filename
+  )
+}
+
+function fileHandlerFactory (fastify: FastifyInstance) {
```
jean-michelet, line 24:
```suggestion
function createFileHandler (fastify: FastifyInstance) {
```

src/routes/api/tasks/index.ts
```diff
@@ -246,24 +244,21 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
 
         const filename = `${id}_${file.filename}`
 
-        const affectedRows = await trx<Task>('tasks')
-          .where({ id })
-          .update({ filename })
+        const task = await trx<Task>('tasks').where({ id }).first()
```
jean-michelet, line 247:
```suggestion
        const task = await trx<Task>('tasks').select('filename').where({ id }).first()
```

test/routes/api/tasks/tasks.test.ts
```diff
@@ -509,6 +513,31 @@ describe('Tasks api (logged user only)', () => {
...
+          payload: form,
+          headers: form.getHeaders()
+        })
+
+        assert.strictEqual(res.statusCode, 200)
+
+        const { message } = JSON.parse(res.payload)
+        assert.strictEqual(message, 'File uploaded successfully')
+
+        assert.strictEqual(mockUnlink.callCount(), 1)
```
jean-michelet, line 538: A safer way to test this would be to upload the first time with a different image (so different filename):
```ts
await uploadImageForTask(
  app, 
  taskId, 
  imageToReplacePath,  // Here...
  uploadDirTask
 )
```

And check after the request that the file is not present in the upload dir anymore.

jean-michelet (approved):
Take a look at my feedbacks plz.

Besides that, LGTM!

jean-michelet:
Can I merge @climba03003 @Fdawgs?

climba03003 (changes requested):
I would go for temporary file approach.
1. Save the file with random name.
2. Remove the old file.
3. Move the tmp file to proper location.

It can ensure when the transaction is rollback or file upload is failed.
You still have the copy of old file.

jean-michelet:
Do you plan to terminate this PR @Turn0xx? Or should I terminate it for you?

Turn0xx:
@jean-michelet I plan to finish it, but i don't have much time this weeks so i don't know when.

jean-michelet:
Fixed by #93

Issue #52, opened 2024-11-02
Improve architecture


jean-michelet:
Currently, we have a couple of problems:
1. Controllers that handle business layer as well as data layer
2. Plugins lack clear modularity, making it harder to distinguish between core infrastructure (like db) and domain-specific logic.
3. The structure is organized by technical concerns (e.g., routes, plugins) rather than by business domains, which can make it challenging to locate and manage related functionalities as the codebase grows.

It is ok for a very small demo, but it becomes inconvenient to maintain if we add more features.

1. I think controllers should be dedicated to HTTP handling (e.g., request/response handling, parameter validation...).

2. We can rename the `plugins` folder to something more representative, like `infrastructure` or `shared`, to reflect that these are reusable components providing core capabilities across the project.

3. Maybe we should rename `routes` to `domains` or `modules` to represent distinct business areas (e.g., `auth`, `tasks`, etc.).

```
src/
├── core/                        # Core reusable logic
│   ├── authorization.ts
│   └── scrypt.ts
├── domains/                  # Business domains
│   ├── auth/                 
│   │   ├── controllers/      
│   │   ├── services/         
│   │   ├── schemas.ts     
│   │   └── routes.ts        
│   ├── tasks/                
│   │   ├── controllers/   
│   │   ├── services/  
│   │   ├── schemas.ts  
│   │   └── routes.ts 
│   └── ...
└── app.ts     
```

Wdyt? Do you want to work on it?

jean-michelet:
Ping for feedback: @climba03003 @melroy89 @Fdawgs

Fdawgs:
I'd keep it as `routes` and `plugins` as then the language is consistent with how the official Fastify documentation and plugin docs refer to these concepts, which should make it easier to understand.

For example, I have a database book that switches between using tuple, row, record, and entry, and it's a pain to follow along.

Turn0xx:
I fully support this initiative to reorganize the codebase, and i can heavily help on that.
Currently, controllers are handling too much, including business logic, which makes file navigation and code comprehension more challenging.
Also the reorganization by "buisness" domains can be very beneficial, so each component purpose is much clearer, intuitive...

Turn0xx:
> I'd keep it as `routes` and `plugins` as then the language is consistent with how the official Fastify documentation and plugin docs refer to these concepts, which should make it easier to understand.

For `routes`, certainly easier to understand since it's standard/present on most demos.. Imo it is not really representative of the real purpose of the components inside.

[...]

PR #54, opened 2024-11-08
refactor: improve global architecture

Turn0xx:
- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      
      
#52

### DIFF
--- a/src/modules/auth/index.route.ts
+++ b/src/modules/auth/index.route.ts
@@ -1,5 +1,5 @@
 import { FastifyPluginAsyncTypebox, Type } from "@fastify/type-provider-typebox";
-import createAuthController from '../controllers/index.js'
+import createAuthController from './controllers/index.js'
 
 
 
--- a/src/modules/auth/index.route.ts
+++ b/src/modules/auth/index.route.ts
@@ -1,8 +1,5 @@
-import { FastifyPluginAsyncTypebox, Type } from "@fastify/type-provider-typebox";
+import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox'
 import createAuthController from './controllers/index.js'
-
-
-
 
 const authRoutes: FastifyPluginAsyncTypebox = async (fastify) => {
   const authController = createAuthController(fastify)
--- a/src/modules/auth/auth.service.ts
+++ b/src/modules/auth/auth.service.ts
@@ -1,6 +1,5 @@
-import { FastifyInstance } from 'fastify';
-import fp from 'fastify-plugin';
-
+import { FastifyInstance } from 'fastify'
+import fp from 'fastify-plugin'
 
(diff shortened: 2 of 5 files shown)


Turn0xx:
This PR introduces a modular architecture, organized by functional domains. 
Each modules encapsulate its own declarations / definitions.


This is the actual structure that i'm going to follow: 

```
modules
├── accounts
│   ├── index.route.ts
├── auth
│   ├── auth.service.ts
│   ├── controllers
│   │   ├── index.ts
│   │   └── login.controller.ts
│   ├── index.route.ts
└── tasks
       ├── index.route.ts
```

Each `module` is going to have 
 - A bunch of controllers (generally one controller per route), who consumes services through the fastify instance.
 - One or more services, implemented as plugins.
 - Route file that serve as the api entry points
 
 Wdyt on this approach ?

Turn0xx:
@jean-michelet You mean to much for the auth module or for all modules ? yeah for auth module just an index file would do the work. However i think for other modules like tasks, we can create multiple controller files per action area.

jean-michelet:
Are you sure it will not break ts inference? One file by controller function is too much I think.

Turn0xx:
> Are you sure it will not break ts inference? 
What do you mean by ts inference ?

> One file by controller function is too much I think.
It's not really one file by controller function but one file by action area.
But for this demo you are right there is not really a lot of actions area per module so i'm going to make just 1 file.

And the same thing for services i think.

src/modules/auth/index.route.ts
```diff
@@ -0,0 +1,9 @@
+import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox'
+import createAuthController from './controllers/index.js'
+
+const authRoutes: FastifyPluginAsyncTypebox = async (fastify) => {
+  const authController = createAuthController(fastify)
+  fastify.post('', {}, authController.login)
```
mcollina, line 6: This is an antipattern. You are splitting the route code from its definitions, and you shouldn't really do that at all, as they are tightly coupled together.

mcollina (changes requested):

jean-michelet:
I think it would be better to create a new repository using an IoC container to provide examples of how to implement a hexagonal architecture with Fastify. Something using the [awilix](https://github.com/fastify/fastify-awilix) plugin. I could even create a boilerplate that simplifies container construction and allows overriding services for testing purposes.  

This demo is intended for a broad audience that doesn’t necessarily master or have an interest in these practices.

jean-michelet:
Close due to inactivity.

Issue #51, opened 2024-10-29
Strengthens payload validation


jean-michelet:
Currently, our schemas are very permissive on what can be sent and tests are missing.

~As payload validation logic can be very fine-grained, I recommend combining unit tests to precisely validate schema constraints and e2e tests to validate that our endpoints support validation.~. Edit, unit tests for dtos seems overkill for the scope of this demo.

Mathieuka:
Still relevant ? 

@jean-michelet I'd like to take it, I need more context thanks

jean-michelet:
If you look at the issue mention above your comment.

PR #68, opened 2024-12-15
feat: improve schemas configuration and test coverage

jean-michelet:
Related to #51 

The goal of this pr is to strengthens payload validation.
I added e2e tests to ensure payload validation on POST/PATCH.
I also fixed a design issue, `author_id` was sent as part of the task creation payload, but I think it should be guessed from the session.


jean-michelet:
Security Hotspot is a false positive.

climba03003 (approved):

PR #47, opened 2024-10-27
feat: Add SonarQube scan to CI workflow

Heyner128:
Fixes #42 

I've added a step in the CI workflow of Github Actions to scan the new code with SonarQube.
Due to the nature of the action, one secret and three variables need to be added to the repository:

Variables: 
- SONAR_ORGANIZATION => the name of the organization 
- SONAR_PROJECT_KEY => the project key
- SONAR_HOST_URL => the SonarCloud URL e.g. https://sonarcloud.io

Secrets:
- SONAR_TOKEN => the API token for SonarCloud

Also the automatic analysis option needs to be disabled as it conflicts with the CI workflows ([documentation](https://docs.sonarsource.com/sonarcloud/advanced-setup/automatic-analysis/#conflict-with-ci-based-analysis)) in the SonarCloud project administration settings in *Administration > Analysis method > Automatic Analysis*

You can check a working execution in this [link](https://github.com/Heyner128/demo-fastify/actions/runs/11540957444)

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #76, opened 2025-01-05
Sonar cloud ignore

jean-michelet:
I think we should remove `test` folder from the analysis as we have a bunch of false positives.
@Eomm had done it on the dashboard [here](https://github.com/fastify/demo/pull/47#issuecomment-2460634652), but it does not seem to work probably because not recursive: `test/*` should be `test/**/*`.

Anyway, I think it's better to integrate scanner configuration into the codebase.


.sonarcloud.properties
```diff
@@ -0,0 +1 @@
+sonar.exclusions=test/**/*,scripts/seed-database.ts
```
jean-michelet, line 1: I added `scripts/seed-database.ts` because of this line: https://github.com/fastify/demo/blob/c29c4ed6c8a7d7e8d2ea98a00b8dfef489ecbf2d/scripts/seed-database.ts#L51

vincent89291 (approved):

gurgunday (approved):
👍

vincent89291 (approved):

Issue #82, opened 2025-02-27
Is the root parameter reasonable?


guchenhui:
https://github.com/fastify/demo/blob/ac5312616294e24ec00f58555c495af31200b8d5/src/plugins/external/static.ts#L18

Why is the root directory of the project static?

Shouldn't it be written like this

```
root: path.join(import.meta.dirname, '../../..', fastify.config.UPLOAD_DIRNAME),
```

jean-michelet:
I think it is better yep, even if prefix is set.

We can rename `dirPath` to `root`.  

I would also add a sanitization check for `UPLOAD_DIRNAME` (no empty string, no relative path) to throw an error in case of a misconfiguration at boot.

Do you want to push a PR for this.

jean-michelet:
Thx @guchenhui

PR #83, opened 2025-02-27
make root parameters more reasonable and strict

guchenhui:
#82

### DIFF
--- a/src/plugins/external/env.ts
+++ b/src/plugins/external/env.ts
@@ -70,6 +70,8 @@ const schema = {
     // Files
     UPLOAD_DIRNAME: {
       type: 'string',
+      minLength: 1,
+      pattern: '^(?!.*\\.{2}).*$',
       default: 'uploads'
     },
     UPLOAD_TASKS_DIRNAME: {
--- a/test/app/env.test.ts
+++ b/test/app/env.test.ts
@@ -15,14 +15,15 @@ it('UPLOAD_DIRNAME should not contain ..', async (t) => {
       await fastify.close()
     })
   }
+})
 
-  const successPath = [UPLOAD_DIRNAME.default]
-  for (let i = 0; i < successPath.length; i++) {
-    const fastify = Fastify()
-    await assert.doesNotReject(async () => {
-      await fastify.register(Env, { confKey, dotenv: false, data: {}, schema: { type, properties: { UPLOAD_DIRNAME: { ...UPLOAD_DIRNAME, default: successPath[i] } } } })
-    }).finally(async () => {
-      await fastify.close()
-    })
-  }
+it('UPLOAD_DIRNAME.default should be a valid dirname', async (t) => {
+  const { confKey, schema: { type, properties: { UPLOAD_DIRNAME } } } = autoConfig
+
+  const fastify = Fastify()
+  await assert.doesNotReject(async () => {
+    await fastify.register(Env, { confKey, dotenv: false, data: {}, schema: { type, properties: { UPLOAD_DIRNAME: { ...UPLOAD_DIRNAME, default: UPLOAD_DIRNAME.default } } } })
+  }).finally(async () => {
+    await fastify.close()
+  })
 })
--- a/test/app/env.test.ts
+++ b/test/app/env.test.ts
@@ -4,40 +4,27 @@ import assert from 'node:assert'
 import Env, { autoConfig } from '../../src/plugins/external/env.js'
 
 it('UPLOAD_DIRNAME should not contain ..', async (t) => {
-  const { schema: { type, properties: { UPLOAD_DIRNAME } }, ...otherAutoConfig } = autoConfig
+  const { confKey, schema: { type, properties: { UPLOAD_DIRNAME } } } = autoConfig
 
-  const failPath = ['/../', '../', '/..', `${UPLOAD_DIRNAME.default}/../${UPLOAD_DIRNAME.default}`, `../${UPLOAD_DIRNAME.default}`, `${UPLOAD_DIRNAME.default}/..`]
+  const failPath = ['/../', '../', '/..']
   for (let i = 0; i < failPath.length; i++) {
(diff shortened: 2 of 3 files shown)


src/plugins/external/env.ts
```diff
@@ -70,6 +70,8 @@ const schema = {
     // Files
     UPLOAD_DIRNAME: {
       type: 'string',
+      minLength: 1,
+      pattern: '^(?!.*\\.{2}).*$',
```
jean-michelet, line 74: Should we add an integration test?

test/app/env.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const fastify = Fastify()
+    fastify.register(Env, { ...otherAutoConfig, schema: { type, properties: { UPLOAD_DIRNAME } } })
+
+    try {
+      await fastify.ready()
+    } catch (error) {
+      assert.fail('should succeed')
+    } finally {
+      await fastify.close()
+    }
```
jean-michelet, line 41: Use [assert.doesNotReject](https://nodejs.org/api/assert.html#assertdoesnotrejectasyncfn-error-message)

test/app/env.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import Env, { autoConfig } from '../../src/plugins/external/env.js'
+
+it('UPLOAD_DIRNAME should not contain ..', async (t) => {
+  const { schema: { type, properties: { UPLOAD_DIRNAME } }, ...otherAutoConfig } = autoConfig
+
+  const failPath = ['/../', '../', '/..', `${UPLOAD_DIRNAME.default}/../${UPLOAD_DIRNAME.default}`, `../${UPLOAD_DIRNAME.default}`, `${UPLOAD_DIRNAME.default}/..`]
+  for (let i = 0; i < failPath.length; i++) {
+    process.env.UPLOAD_DIRNAME = failPath[i]
+    const fastify = Fastify()
+    fastify.register(Env, { ...otherAutoConfig, schema: { type, properties: { UPLOAD_DIRNAME } } })
```
jean-michelet, line 13: We should not pollute process.env, it can causes several issues.
Maybe just override the UPLOAD_DIRNAME.default value with the invalid path when registering the plugin?

jean-michelet:
```ts
const invalidSchema = {
  type,
  properties: {
    ...properties,
    UPLOAD_DIRNAME: {
      ...properties.UPLOAD_DIRNAME,
      default: failPath
    },
  },
};

fastify.register(Env, { ...otherAutoConfig, schema: invalidSchema });
```

test/app/env.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+it('UPLOAD_DIRNAME should not contain ..', async (t) => {
+  const { confKey, schema: { type, properties: { UPLOAD_DIRNAME } } } = autoConfig
+
+  const failPath = ['/../', '../', '/..']
+  for (let i = 0; i < failPath.length; i++) {
+    const fastify = Fastify()
+    await assert.rejects(async () => {
+      await fastify.register(Env, { confKey, dotenv: false, data: {}, schema: { type, properties: { UPLOAD_DIRNAME: { ...UPLOAD_DIRNAME, default: failPath[i] } } } })
+      await fastify.ready()
```
jean-michelet, line 14: You don't need to put everything in the tested callback, `fastify.ready` is enough

test/app/env.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      await fastify.close()
+    })
+  }
+
+  const successPath = [UPLOAD_DIRNAME.default]
+  for (let i = 0; i < successPath.length; i++) {
+    const fastify = Fastify()
+    await assert.doesNotReject(async () => {
+      await fastify.register(Env, { confKey, dotenv: false, data: {}, schema: { type, properties: { UPLOAD_DIRNAME: { ...UPLOAD_DIRNAME, default: successPath[i] } } } })
+      await fastify.ready()
```
jean-michelet, line 25: same

test/app/env.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    await assert.rejects(async () => {
+      await fastify.register(Env, { confKey, dotenv: false, data: {}, schema: { type, properties: { UPLOAD_DIRNAME: { ...UPLOAD_DIRNAME, default: failPath[i] } } } })
+      await fastify.ready()
+    }, { message: `env/UPLOAD_DIRNAME must match pattern "${UPLOAD_DIRNAME.pattern}"` }).finally(async () => {
+      await fastify.close()
+    })
+  }
+
+  const successPath = [UPLOAD_DIRNAME.default]
+  for (let i = 0; i < successPath.length; i++) {
```
jean-michelet, line 21: If there is only one item in the array, we don't need to perform an iteration.
I would put the success assertion in an other test:
```ts
it('UPLOAD_DIRNAME.default should be a valid dirname', async (t) => { 
 // The test case...
})

jean-michelet (approved):

jean-michelet:
Plz, just fix my feedback and it's ok :+1:

test/app/env.test.ts
guchenhui: I tested it and the code that actually works is `await fastify.register()`, even this code `await fastify.ready()` is unnecessary. In order to make the testing case process more pure, i removed `await fastify.ready()`

jean-michelet: This is because you `await` the registration, you generally don't have to but I am not against it.

src/plugins/external/env.ts
gurgunday: What's the goal of this change?

jean-michelet: Configuration mistakes that could include relative path and impact `fastifyStatic` `root` upload file value.

gurgunday (approved):
lgtm

PR #93, opened 2025-03-30
Refactorisation

jean-michelet:
### Encapsulate more logic into utility plugins
Our controllers are currently handling too many responsibilities: database queries, file uploads, and more. IMO, they should focus only on handling HTTP concerns. I made some improvements in this direction, but there's still work to be done. I believe introducing a service layer or a set of use-cases would help to achieve this separation.

### Better logging
I think relying on the `reply.internalServerError()` utility from `@fastify/sensible` is not very useful, and we were essentially blind to the actual reasons why transactions were failing (credits to @fraxken). If we don't implement fine-grained logging, it's better to let errors propagate naturally. The default error handler already does a good job of adding context to thrown errors.

### Hotfixes
I also took the opportunity to address #38 by implementing the temporary file approach recommended by @climba03003 [here](https://github.com/fastify/demo/pull/50#pullrequestreview-2428510084).

## Future improvements
- Encapsulate business logic into dedicated plugins (service objects or use-case decorators). Controllers would no longer handle file management or direct repository calls.
- Improve the folder structure to better reflect which plugins belong to infrastructure, application, shared utilities, or business logic.
- Reduce the use of module augmentation by using `getDecorator<T>` api on next minor, see https://github.com/fastify/fastify/pull/5768


gurgunday (approved):

jean-michelet:
@fastify/collaborators 
Can an additional contributor take a look plz?

metcoder95 (approved):

PR #100, opened 2025-04-05
Upgrade dependencies

jean-michelet:
@Fdawgs 
`dependabot` doesn't seems to be configured very well (or there is an issue with this repo), I would prefer to have PR on a regular basis to upgrade patches and minors.

Because you know better CI, what do you recommend?


Fdawgs:
@jean-michelet because the packages are using the caret (`^`), Dependabot will only make PRs for major releases outside of the specified range (excluding 0.x, where any change is considered breaking), or when a dependency has a CVE against a specific version.

If you want it to make PRs for minor and hotfix/patch releases you'll need to remove the caret, or include a package-lock.json.

jean-michelet:
Ok, I'll update that, we have 100% test coverage, so nothing should break.

Issue #99, opened 2025-04-05
Reduce the use of module augmentation and improve runtime safety by using `getDecorator<T>` api


jean-michelet:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Wait for #93 and Fastify next minor `5.3`.

jean-michelet:
Ready.

PR #102, opened 2025-04-15
Refactor/leverage getdecorator api

jean-michelet:
Implement #99

This change uses the new `getDecorator` API to reduce module augmentation and improve both type safety and runtime safety.
Reference: https://fastify.dev/docs/latest/Reference/Decorators/#getdecoratort-api

There are other ways to ensure runtime safety with decorators, but using TypeScript, this approach it easier as perform static and runtime checks in one place.

I used Symbol to name decorators, don't know what you think about that.


jean-michelet:
Started from an existing branch before pulling, which caused this commit history mess.
Will squash commits properly before merging.

jean-michelet:
I need several feedback on the style, this is why I ask for 3 reviews.

src/plugins/app/authorization.ts
```diff
@@ -1,26 +1,28 @@
...
-    isAdmin: typeof isAdmin;
-  }
-}
+export type AuthorizationManager = ReturnType<typeof createChecker>
+export const kAuthorizationManager = Symbol('app.authorizationManager')
 
-function verifyAccess (this: FastifyRequest, reply: FastifyReply, role: string) {
-  if (!this.session.user.roles.includes(role)) {
-    reply.status(403).send('You are not authorized to access this resource.')
+function createChecker () {
```
jean-michelet, line 9:
```suggestion
function createManager () {
```

src/plugins/app/authentication.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { auth } = this.getDecorator<SessionData>('session')
+  if (auth === undefined) {
+    return false
+  }
+
+  // Instead of accessing the session directly in your application,
+  // you should decorate the request.
+  // This reduces coupling with the authentication strategy.
+  // If you change or add new authentication strategies in the future,
+  // `request.auth` remains the single source of truth.
```
jean-michelet, line 20:
```suggestion
  // `request[kAuth]` remains the single source of truth.
```

jean-michelet:
@Eomm @mcollina @gurgunday 
Not interested in this feature?

I understand that the official demo needs to represent the current state of Fastify's practices.
But it can also be a way for people to discover new features and ways to enjoy Fastify.

Maybe I can add some comments to explain the alternatives to `getDecorator` api? Like module augmentation?

Issue #119, opened 2025-09-17
Cannot use fastify-cli with --options


nicolas-goudry:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.4.0

### Node.js version

20.17.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

NixOS 24.11

### Description

Hi,

First of, thanks for this demo repository, it's filled with goodies I already used and even more I wasn't aware of.

I based one of my projects on this repository and found out that using `fastify-cli` with the `--options` flag ends up in an error I can't quite figure out:

```
TypeError: ajv.validate is not a function
    at envSchema (/home/nicolas/work/perso/demo/node_modules/env-schema/index.js:83:21)
    at fastifyEnv (/home/nicolas/work/perso/demo/node_modules/@fastify/env/index.js:8:20)
    at Plugin.exec (/home/nicolas/work/perso/demo/node_modules/avvio/lib/plugin.js:125:28)
    at Boot._loadPlugin (/home/nicolas/work/perso/demo/node_modules/avvio/boot.js:432:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

To reproduce this error, run the following command:

```bash
npm run dev:start -- --options
```

From what I could gather, this issue is related to [this part of the code](https://github.com/fastify/demo/blob/1d430da06ad4df398d9b390b5f425d753ab06a6b/src/app.ts#L10-L15) in `src/app.ts`:

```js
export const options = {
  ajv: {
    customOptions: {
      coerceTypes: 'array',
      removeAdditional: 'all'
    }
  }
}
```

If removing the `ajv` key entirely, it works again. I tried adding other options like `ignoreDuplicateSlashes` and such and they do work.

I’d love to be able to fix this, but I don't really know where to start.

I can deduce from the error message that this has something to do with the `@fastify/env` plugin and its dependency `env-schema`. I’ll try to look into it if I have some time.

### Link to code that reproduces the bug

https://github.com/fastify/demo

### Expected Behavior

The `--options` flag of `fastify-cli` should use custom options defined in the app service without any errors.

nicolas-goudry:
I looked through the `env-schema` code and it seems to be quite a subtle bug.

---

**Why is this happening?**

To understand why this happens, we have to look at how `fastify-cli` handles the `options` export from plugins (which `serviceApp` is).

[...]

PR #121, opened 2025-09-19
fix: stop propagating options

jean-michelet:
Can lead to issues with Core plugins expecting `options` to have a certain shape and key/values.

See #119


is2ei (approved):
LGTM!

gurgunday (approved):

PR #133, opened 2025-12-22
Docs: add env setup before db commands

lraveri:
Fixes #132 document copying `.env.example` to `.env` before running db scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,16 @@ Install the dependencies:
 npm install
 ```
 
-### Database
+### Environment variables
 
-Prepare the environment file:
+Create a `.env` file based on `.env.example` and update values as needed.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,8 +18,18 @@ Install the dependencies:
 npm install
 ```
 
+### Environment variables
+
+Create a `.env` file based on `.env.example` and update values as needed.
+
+Make sure `COOKIE_SECRET` is set to a 32-byte secret (64 hex chars) in your `.env`. You can generate a random secret with:
```
jean-michelet:
```suggestion
Make sure `COOKIE_SECRET` is set to a secret with at least 32 characters.
You can generate a random secret with:
```

lraveri: I’ve pushed a new commit to update the README as you suggested

jean-michelet (approved):

PR #147, opened 2026-03-30
chore: update @fastify/env dependency

Tony133:
Proposal:

- update `@fastify/env` dependency to `v6.0.0`, because the `v5.0.3` it still uses `dotenv` internally


gurgunday (approved):
Lgtm

PR #167, opened 2026-09-13
Remove temp file when file is successfully uploaded

aaronofinn:
When uploading a file for a task that already has a file, it moves the old file to a temp directory. But on a successful upload, the temporary file still stays there so this is just a small fix to remove that temp file when the operation is successful.

### How to reproduce (before this fix)

```bash
curl -c /tmp/c.txt -X POST http://localhost:3000/api/auth/login \
  -H "Content-Type: application/json" \
  -d '{"email":"basic@example.com","password":"Password123$"}'

# upload once
curl -b /tmp/c.txt -X POST http://localhost:3000/api/tasks/1/upload -F "file=@/path/to/image.png"
# upload again onto the same task
curl -b /tmp/c.txt -X POST http://localhost:3000/api/tasks/1/upload -F "file=@/path/to/image.png"

ls uploads/tasks/temp/
# a leftover temp-<random>-1_image.png sits here after the second upload succeeded
```

#### Checklist


jean-michelet:
I think we should remove this feature of file upload, I tried to do this to illustrate the use of the multipart plugin and streams usage, but there are several system design flaws that fixing theme would make to much noise for a basic demo.

PR #165, opened 2026-09-13
Remove stale references to standalone along with concurrently dev dependency

aaronofinn:
Removing stale referencing to standalone.

Standalone code was originally removed as part of #129 but I think some of the documentation was left stale, so this is a PR to just clean that up.

#### Checklist

- [X] run `npm run test && npm run benchmark --if-present`
- [ ] tests and/or benchmarks are included
- [X] documentation is changed or added
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #168, opened 2026-09-13
Fix flaky test

aaronofinn:
#### Description

`tasksRepository.paginate` sorts by `.orderBy('created_at', q.order)` alone. `created_at` is a `TIMESTAMP` column with 1-second resolution, so tasks created within the same second (seeded, or created back to back) tie exactly, and with no secondary sort key, which row lands at a given page/offset is undefined, it can vary between identical requests.

#### How this was found

[CI failed](https://github.com/fastify/demo/actions/runs/34774603376/job/103771236580) on an unrelated change in a recent PR: #165 with `test/routes/api/tasks/tasks.test.ts:136`, "should paginate by page and limit", expecting `'Task 2'` and getting `'Task 4'`. Reproduced locally, running the same suite repeatedly against unmodified `main` gives different results across runs with no code changes, confirming genuine nondeterminism rather than a one-off. 

Confirmed the test passed after a retry and is therefore flaky.

So the fix is instead to tie the order by the id as it will always be unique and thus, in order.

#### Checklist


jean-michelet (approved):

aaronofinn:
@jean-michelet would you mind merging? I don't have priveleges for some reason

PR #169, opened 2026-09-13
Add sameSite: 'lax' to cookie settings

aaronofinn:
#### Description

`session.ts` sets `secure` and `httpOnly` on the session cookie but never `sameSite` and there's no CSRF token mechanism anywhere in the app either. Whatever protection exists today is Chrome's own implicit default (treating a missing `SameSite` as `Lax`), an accident of the browser, not something this app's config actually asserts, and not something every client is guaranteed to apply consistently.

Adding sameSite: 'lax' to the cookie settings allows us to explicitly enforce that the session cookie is only sent on same-site requests, rather than depending on whichever browser happens to be in use applying its own undocumented default when the attribute is missing.

#### Checklist


jean-michelet:
> Adding sameSite: 'lax' to the cookie settings allows us to explicitly enforce that the session cookie is only sent on same-site requests

That's Strict, not Lax.

I planed to introduce CSRF building a SPA, but I always failed to fastify/vite with React.
Let's avoid it for now imo.

jean-michelet (approved):

aaronofinn:
@jean-michelet would you mind merging? I don't have priveleges for some reason

PR #170, opened 2026-09-13
Regenerate the session before logging in as a new user

aaronofinn:
#### Description

`auth/index.ts`'s login handler assigns `request.session.user = {...}` and calls `request.session.save()`, but never rotates the session ID. If a valid session cookie is already present when `/login` is called, `@fastify/session` reuses that same ID, it just overwrites who the session belongs to.

If an attacker could get a session ID they already know onto a victim's browser before the victim authenticates (however that happens), the ID staying the same across login means the attacker's foreknowledge of it remains useful after the victim logs in. Regenerating the session ID at the moment authentication state changes is the standard defense, it invalidates whatever ID existed beforehand.

#### Checklist


jean-michelet:
We need this: https://github.com/fastify/session/pull/348
@fastify/release: possible to release a patch of fastify/session plz?

Tony133:
> We need this: [fastify/session#348](https://github.com/fastify/session/pull/348) @fastify/release: possible to release a patch of fastify/session plz?

Done! see new release here: https://github.com/fastify/session/releases/tag/v11.1.3 🔥 👀

jean-michelet:
@Tony133 Thx!
@aaronofinn Can you plz update the version?

aaronofinn:
@jean-michelet done, can you merge when ready? I don't have access

.npmrc
```diff
@@ -1,2 +1,3 @@
 ignore-scripts=true
 min-release-age=7
+min-release-age-exclude=@fastify/session
```
jean-michelet: Sorry, we're gonna wait.

